PR #6, opened 2021-05-07
Don't delete node pools for imported clusters

cmurphy:
Also fix examples for imported clusters.

### DIFF
--- a/controller/aks-cluster-config-handler.go
+++ b/controller/aks-cluster-config-handler.go
@@ -622,58 +622,60 @@ func (h *Handler) updateUpstreamClusterState(ctx context.Context, secretsCache w
 		}
 	}
 
-	agentPoolClient, err := aks.NewAgentPoolClient(credentials)
-	if err != nil {
-		return config, err
-	}
-
-	downstreamNodePools, err := utils.BuildNodePoolMap(config.Spec.NodePools, config.Spec.ClusterName)
-	if err != nil {
-		return config, err
-	}
-
-	// check for updated NodePools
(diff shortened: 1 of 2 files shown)


controller/aks-cluster-config-handler.go
```diff
@@ -627,53 +627,55 @@ func (h *Handler) updateUpstreamClusterState(ctx context.Context, secretsCache w
 		return config, err
 	}
 
-	downstreamNodePools, err := utils.BuildNodePoolMap(config.Spec.NodePools, config.Spec.ClusterName)
-	if err != nil {
-		return config, err
-	}
+	if config.Spec.NodePools != nil {
```
thedadams, line 630: nit: put the `aks.NewAgentPoolClient` call above in this if-block because it isn't needed otherwise.

thedadams (dismissed):

thedadams (approved):

rmweir (commented):
To clarify this is more, don't delete node pools if nodepools field is not managed, correct? So if you do have an imported cluster but you've edited the nodepools then upstream node pool not included in spec will be deleted? If this all is correct then LGTM.

cmurphy:
> To clarify this is more, don't delete node pools if nodepools field is not managed, correct?

Yes, that's correct.

cbron (approved):

cbron:
@cmurphy can you add an issue to this PR

cmurphy:
Issue: https://github.com/rancher/rancher/issues/32734

PR #7, opened 2021-05-07
Fix error messages

cmurphy:
`err` is nil at both of these points, so it's unnecessary to log it. Use
the cluster status instead.


thedadams (approved):

mjura (approved):
LGTM

cbron (approved):

PR #8, opened 2021-05-07
Fix VNet subnet ID builder

cmurphy:
Without this change, the subnet ID renders the VirtualNetwork and Subnet
pointer values as hexadecimal values instead of the strings they
represent, and the virtual network cannot be found.


thedadams (approved):

mjura (approved):
LGTM

cbron (approved):

PR #10, opened 2021-05-12
Add AKS chart and CI

cmurphy:
https://github.com/rancher/rancher/issues/32293


charts/aks-operator/values.yaml
```diff
@@ -0,0 +1,11 @@
+global:
+  systemDefaultRegistry: ""
+
+aksOperator:
+  image:
+    repository: rancher/aks-operator
+    tag: v1.0.1-rc1
```
thedadams, line 7: I think it's better to set this to a nonsense value to help indicate that it is overwritten. Something like `999` or `v0.0.0`.

thedadams (commented):
One minor change.

charts/aks-operator/values.yaml
cmurphy: done

thedadams (approved):

mjura (approved):
LGTM

cbron (approved):
lgtm, but may need to wait for build permissions to tag it.

PR #11, opened 2021-05-13
Add HTTP Application Routing

mjura:
Add new option to addon profile HTTPApplicationRouting which
configures an Ingress controller in AKS cluster.

Issue: https://github.com/rancher/rancher/issues/32674

### DIFF
--- a/controller/aks-cluster-config-handler.go
+++ b/controller/aks-cluster-config-handler.go
@@ -581,6 +581,12 @@ func BuildUpstreamClusterState(ctx context.Context, secretsCache wranglerv1.Secr
 		upstreamSpec.LinuxSSHPublicKey = sshKeys[0].KeyData
 	}
 
+	// set addons profile
+	addonProfile := clusterState.AddonProfiles
+	if addonProfile != nil && addonProfile["httpApplicationRouting"] != nil {
+		upstreamSpec.HTTPApplicationRouting = addonProfile["httpApplicationRouting"].Enabled
+	}
+
 	// set API server access profile
 	upstreamSpec.PrivateCluster = to.BoolPtr(false)
 	if clusterState.APIServerAccessProfile != nil {
@@ -597,7 +603,8 @@ func BuildUpstreamClusterState(ctx context.Context, secretsCache wranglerv1.Secr
 
 // updateUpstreamClusterState compares the upstream spec with the config spec, then updates the upstream AKS cluster to
 // match the config spec. Function returns after a update is finished.
-func (h *Handler) updateUpstreamClusterState(ctx context.Context, secretsCache wranglerv1.SecretCache, config *aksv1.AKSClusterConfig, upstreamSpec *aksv1.AKSClusterConfigSpec) (*aksv1.AKSClusterConfig, error) {
+func (h *Handler) updateUpstreamClusterState(ctx context.Context, secretsCache wranglerv1.SecretCache,
+	config *aksv1.AKSClusterConfig, upstreamSpec *aksv1.AKSClusterConfigSpec) (*aksv1.AKSClusterConfig, error) {
 	credentials, err := aks.GetSecrets(secretsCache, &config.Spec)
 	if err != nil {
 		return config, err
@@ -697,6 +704,14 @@ func (h *Handler) updateUpstreamClusterState(ctx context.Context, secretsCache w
 		}
 	}
 
+	// check addon HTTP Application Routing
+	if config.Spec.HTTPApplicationRouting != nil {
+		if to.Bool(config.Spec.HTTPApplicationRouting) != to.Bool(upstreamSpec.HTTPApplicationRouting) {
+			logrus.Infof("Updating HTTP application routing for cluster [%s]", config.Spec.ClusterName)
+			updateAksCluster = true
+		}
+	}
+
 	if updateAksCluster {
 		resourceGroupsClient, err := aks.NewResourceGroupClient(credentials)
 		if err != nil {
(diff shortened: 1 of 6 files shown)


internal/aks/create.go
```diff
@@ -128,6 +130,15 @@ func CreateOrUpdateCluster(ctx context.Context, cred *Credentials, clusterClient
 		}
 	}
 
+	addonProfiles := map[string]*containerservice.ManagedClusterAddonProfile{
+		"httpApplicationRouting": {
+			Enabled: spec.HTTPApplicationRouting,
+		},
+	}
+	if !hasHTTPApplicationRoutingSupport(spec) {
+		delete(addonProfiles, "httpApplicationRouting")
```
cmurphy, line 139: Why delete after it's added? Why not
```suggestion
	if hasHTTPApplicationRoutingSupport(spec) {
		addonProfiles := map[string]*containerservice.ManagedClusterAddonProfile{
			"httpApplicationRouting": {
				Enabled: spec.HTTPApplicationRouting,
			},
		}
```

internal/aks/create.go
```diff
@@ -128,6 +130,15 @@ func CreateOrUpdateCluster(ctx context.Context, cred *Credentials, clusterClient
 		}
 	}
 
+	addonProfiles := map[string]*containerservice.ManagedClusterAddonProfile{
+		"httpApplicationRouting": {
+			Enabled: spec.HTTPApplicationRouting,
+		},
+	}
+	if !hasHTTPApplicationRoutingSupport(spec) {
```
cmurphy, line 138: some kind of log message might be nice since this field is just being ignored

cmurphy (changes requested):

internal/aks/create.go
mjura: good point

internal/aks/create.go
mjura: As I see this is also for China Azure cloud, probably this should be general information about potential limitation for China aks deployments.

cmurphy (dismissed):

rmweir (dismissed):
LGTM

cmurphy (approved):

rmweir (approved):

controller/aks-cluster-config-handler.go
```diff
@@ -593,6 +593,12 @@ func BuildUpstreamClusterState(ctx context.Context, secretsCache wranglerv1.Secr
 		upstreamSpec.WindowsAdminPassword = windowsProfile.AdminPassword
 	}
 
+	// set addons profile
+	addonProfile := clusterState.AddonProfiles
+	if addonProfile != nil {
```
thedadams, line 598:
```suggestion
	if addonProfile["httpApplicationRouting"] != nil {
```

thedadams (changes requested):

controller/aks-cluster-config-handler.go
mjura: Added this suggestion

controller/aks-cluster-config-handler.go
```diff
@@ -593,6 +593,12 @@ func BuildUpstreamClusterState(ctx context.Context, secretsCache wranglerv1.Secr
 		upstreamSpec.WindowsAdminPassword = windowsProfile.AdminPassword
 	}
 
+	// set addons profile
+	addonProfile := clusterState.AddonProfiles
```
rmweir, line 597: Would this ever be nil?

cmurphy (approved):

internal/aks/create.go
```diff
@@ -151,7 +163,7 @@ func CreateOrUpdateCluster(ctx context.Context, cred *Credentials, clusterClient
 			AuthorizedIPRanges: spec.AuthorizedIPRanges,
 		}
 	}
-	if spec.PrivateCluster != nil && *spec.PrivateCluster {
+	if to.Bool(spec.PrivateCluster) {
```
rmweir, line 166: If PrivateCluster is nil, will this cause it to be overwritten with false in upstream if it is true?

rmweir (changes requested):

internal/aks/create.go
thedadams: Good catch!

controller/aks-cluster-config-handler.go
mjura: I have added additional check to verify, when it might be `nil`

internal/aks/create.go
mjura: Yeap, that's right

cmurphy (approved):

controller/aks-cluster-config-handler.go
thedadams: It's fine to have it, but you don't need it: reading from a `nil` map is valid in Go.

thedadams (dismissed):

internal/aks/create.go
rmweir: Is there any logic for disabling this if privateCluster is set to false in the spec but it is currently enabled upstream?

mjura: yes, it is

mjura: this is simple change which is simplifying this `if` condition, but `spec.PrivateCluster != nil && *spec.PrivateCluster` is  equivalent of `to.Bool(spec.PrivateCluster)`

rmweir: But what if spec.PrivateCluster != niil && !*spec.PrivateCluster? How would upstream PrivateCluster be disabled?

rmweir (changes requested):
If not setting PrivateCluster causes it to disable in upstream, LGTM.

internal/aks/create.go
mjura: After enabling `PrivateCluster` it is not possible to switch it off, this is limitation of Azure API. We don't have to worry about it.

thedadams (approved):

cmurphy (approved):

PR #12, opened 2021-05-14
Set norman type to 'nullablestring' for *strings

cmurphy:
Opt into a new field type in norman that will convert `*string`s to
`{nullable: true, type: nullablestring}` in the API schema, and back to
`*string` in the generated client code. Without this, `*string`s get
translated to just `string` in the generated code, which breaks clients'
ability to set fields to nil.

https://github.com/rancher/rancher/issues/32440


mjura (approved):
LGTM

thedadams (approved):

PR #14, opened 2021-05-20
Move TenantID to Cloud Credentials

mjura:


cmurphy (commented):
lgtm but I think needs to wait for tenant ID to be added to the cloud credential/machine driver

internal/aks/client.go
```diff
@@ -94,11 +94,15 @@ func GetSecrets(secretsCache wranglerv1.SecretCache, spec *aksv1.AKSClusterConfi
 		return nil, fmt.Errorf("couldn't find secret [%s] in namespace [%s]", id, ns)
 	}
 
+	tenantIDBytes := secret.Data["azurecredentialConfig-tenantId"]
```
rmweir, line 97: You may want to add this to the example in the README where secret creation is demonstrated.

rmweir (commented):
LGTM, aside from one comment about an addition to be made to README.

thedadams (approved):

internal/aks/client.go
mjura: I have opened new PR https://github.com/rancher/aks-operator/pull/15

PR #15, opened 2021-05-24
Update example files with new options

mjura:


examples/create-example.yaml
```diff
@@ -8,8 +8,10 @@ spec:
   clusterName: "my-cluster"
   baseUrl: "https://management.azure.com/"
   authBaseUrl: "https://login.microsoftonline.com"
-  tenantId: "REPLACE_WITH_TENANT_ID"
   azureCredentialSecret: "REPLACE_WITH_K8S_SECRETS_NAME"
+  privateCluster: false,
+  windowsAdminUsername: "my-admin-user",
+  windowsAdminPassword: "my-secure-passworD",
```
cmurphy, line 14: If you want to include windows in this example then you can remove https://github.com/rancher/aks-operator/blob/master/examples/create-windows-example.yaml

For windows to work you need to set `networkPlugin`, `virtualNetwork`, and `subnet` and the name of the node pool is restricted to 6 characters, `workers2` is too long. This is why I had broken it out into a separate example.

thedadams (approved):

cmurphy (changes requested):

examples/create-example.yaml
mjura: updated

thedadams (dismissed):

examples/create-example.yaml
```diff
@@ -8,8 +8,10 @@ spec:
   clusterName: "my-cluster"
   baseUrl: "https://management.azure.com/"
   authBaseUrl: "https://login.microsoftonline.com"
-  tenantId: "REPLACE_WITH_TENANT_ID"
   azureCredentialSecret: "REPLACE_WITH_K8S_SECRETS_NAME"
+  privateCluster: false,
+  linuxAdminUsername: "rancher-user",
+  linuxAdminPassword: "rancher-password",
```
cmurphy, line 14: There's no `linuxAdminPassword`, it's `linuxSSHPublicKey`

cmurphy (changes requested):

rmweir (approved):
LGTM once Colleen's feedback has been addressed.

examples/create-example.yaml
mjura: ahh yeah, you are right, but there is `sshPublicKey` option

cmurphy (approved):

rmweir (approved):
LGTM

PR #18, opened 2021-05-26
Add additional trusted CAs

thedadams:
When Rancher is installed on a cluster behind a proxy, the operator
should trust the proxy's certificate. It is passed to the deployment via
 a secret the user should setup per the Rancher documentation.

Issue:
https://github.com/rancher/rancher/issues/31846


charts/aks-operator/templates/deployment.yaml
```diff
@@ -25,3 +25,17 @@ spec:
...
+            name: tls-ca-additional-volume
+            subPath: ca-additional.pem
+            readOnly: true
+{{- end }}
+{{- if .Values.additionalTrustedCAs }}
+      volumes:
+        - name: tls-ca-additional-volume
+          secret:
+            defaultMode: 0400
+            secretName: tls-ca-additional
```
jiaqiluo, line 40: Do we need to specify the namespace from where the secret is?

jiaqiluo (changes requested):

charts/aks-operator/templates/deployment.yaml
jiaqiluo: It looks like the api does not support setting the namespace. 
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#secretvolumesource-v1-core

I am not sure if it means we can use the secret in any namespace or only where the pod is in.

jiaqiluo: Oh, I just realized that the aks-operator is installed into the cattle-system namespace where the `tls-ca-additional` secret is created.

thedadams: Yes, that is right. This pattern follows the same one from Rancher: https://github.com/rancher/rancher/blob/master/chart/templates/deployment.yaml#L136-L145

charts/aks-operator/templates/deployment.yaml
```diff
@@ -25,3 +25,15 @@ spec:
           value: {{ .Values.httpsProxy }}
         - name: NO_PROXY
           value: {{ .Values.noProxy }}
+{{- if .Values.additionalTrustedCAs }}
+        volumeMounts:
+          - mountPath: /etc/ssl/certs/ca-additional.pem
+            name: tls-ca-additional-volume
+            subPath: ca-additional.pem
```
jiaqiluo, line 32: The current code will result in creating a file at this full path: 
```
/etc/ssl/certs/ca-additional.pem/ca-additional.pem
```
```suggestion
          - mountPath: /etc/ssl/certs
            name: tls-ca-additional-volume
```

jiaqiluo (changes requested):

charts/aks-operator/templates/deployment.yaml
thedadams: I think the desired result is to create the file.

mjura (approved):
LGTM

jiaqiluo (approved):
LGTM

cbron (approved):

PR #19, opened 2021-05-31
Fix autoscaling for nodedools

mjura:
Fix enabling autoscaling for nodepools after AKS cluster was provisioned

Issue: https://github.com/rancher/rancher/issues/32956

### DIFF
--- a/internal/aks/create.go
+++ b/internal/aks/create.go
@@ -178,7 +178,11 @@ func CreateOrUpdateAgentPool(ctx context.Context, agentPoolClient *containerserv
 		OsType:              containerservice.OSType(np.OsType),
 		VMSize:              containerservice.VMSizeTypes(np.VMSize),
 		Mode:                containerservice.AgentPoolMode(np.Mode),
+		Type:                containerservice.VirtualMachineScaleSets,
 		OrchestratorVersion: np.OrchestratorVersion,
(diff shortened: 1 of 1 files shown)


internal/aks/create.go
```diff
@@ -159,14 +159,22 @@ func CreateOrUpdateCluster(ctx context.Context, cred *Credentials, clusterClient
 }
 
 func CreateOrUpdateAgentPool(ctx context.Context, agentPoolClient *containerservice.AgentPoolsClient, spec *aksv1.AKSClusterConfigSpec, np *aksv1.AKSNodePool) error {
+	if np.EnableAutoScaling == nil {
+		np.EnableAutoScaling = to.BoolPtr(false)
```
thedadams, line 163: Since `np` is a pointer, I think modifying it may cause confusion.

thedadams (changes requested):

rmweir (dismissed):
LGTM

internal/aks/create.go
mjura: Ok, as I have checked it wasn't needed

cmurphy (approved):

rmweir (approved):
LGTM

thedadams (approved):

PR #20, opened 2021-05-31
Add upgrading state for node pools

mjura:
During upgrading node pools there has to be additional phase check


cmurphy (approved):
Is there an issue for this?

thedadams (approved):

rmweir (approved):
LGTM

PR #9, opened 2021-05-11
Add addon profile cluster config

mjura:
Addon profile including two options:
- HTTPApplicationRouting configures an Ingress controller in AKS cluster.
- Monitoring colletcs metrics and logs.

PR #22, opened 2021-06-02
Move packages from internal to pkg

mjura:
AKSv1 needs to reuse CheckLogAnalyticsWorkspaceForMonitoring, which contains
metadata around the region calls. For that all internal packages should be
moved from internal to pkg directory.

Issue: https://github.com/rancher/rancher/issues/32967


cmurphy:
It looks like this ended up being done in https://github.com/rancher/aks-operator/pull/9 after all?

mjura:
It will be done in https://github.com/rancher/aks-operator/pull/9

PR #28, opened 2021-06-10
Add cloud credential validation

mjura:


cmurphy (approved):

controller/aks-cluster-config-handler.go
```diff
@@ -355,10 +355,25 @@ func (h *Handler) validateConfig(config *aksv1.AKSClusterConfig) error {
 		return fmt.Errorf(cannotBeNilError, "azureCredentialSecret", config.ClusterName)
 	}
 
-	if _, err = aks.GetSecrets(h.secretsCache, &config.Spec); err != nil {
+	cred, err := aks.GetSecrets(h.secretsCache, &config.Spec)
+	if err != nil {
 		return fmt.Errorf("couldn't get secret [%s] with error: %v", config.Spec.AzureCredentialSecret, err)
 	}
 
+	cannotBeNilCCError := "field [%s] must be provided for cloud credentials [%s]"
```
thedadams, line 363: nit:
```suggestion
	cannotBeEmptyCCError := "field [%s] must be provided for cloud credentials [%s]"
```

thedadams (commented):
This method containing these changes is only called on cluster create. Should we consider checking the cloud credential on create, import, and update? It is possible that a user could change the cloud credential on a cluster after it is created.

controller/aks-cluster-config-handler.go
mjura: Fixed

cmurphy:
> This method containing these changes is only called on cluster create. Should we consider checking the cloud credential on create, import, and update? It is possible that a user could change the cloud credential on a cluster after it is created.

@thedadams this method is run before moving to the importing phase so I think it's fine there https://github.com/rancher/aks-operator/blob/ff05137c5e97d4d3e26b7f61bb894c193bc829d2/controller/aks-cluster-config-handler.go#L197-L201

I think the point about validating for updates is valid though, @mjura what do you think?

mjura:
@thedadams @cmurphy yes, it is safe to do this in `validateConfig`, because it will be checked before importing and creating phase. I have tested it and works fine

rmweir (approved):
LGTM

thedadams:
@mjura It is possible to change the cloud credential on a cluster. If a user creates an AKS cluster with valid cloud credential and then changes the cloud credential on the cluster to an invalid one, then the updates may fail.

Is this a concern we should address?

mjura:
> @mjura It is possible to change the cloud credential on a cluster. If a user creates an AKS cluster with valid cloud credential and then changes the cloud credential on the cluster to an invalid one, then the updates may fail.
> 
> Is this a concern we should address?

We can add `validateConfig` to function during update phase. I will test it

mjura:
It is not needed any more. It is validated in https://github.com/rancher/aks-operator/blob/master/pkg/aks/client.go#L97
and in Rancher master it is validated also by UI.

cbron:
@mjura was there an issue tracking this, if not please create one and send to test, write up how QA can test.

PR #29, opened 2021-06-11
Improve NetworkProfile configuration

mjura:


pkg/aks/create.go
```diff
@@ -59,25 +69,11 @@ func CreateOrUpdateCluster(ctx context.Context, cred *Credentials, clusterClient
...
-			networkProfile.LoadBalancerSku = loadBalancerSku
-		}
-
-		if spec.NetworkPolicy != nil {
-			networkProfile.NetworkPolicy = containerservice.NetworkPolicy(*spec.NetworkPolicy)
+		networkProfile.PodCidr = spec.NetworkPodCIDR
+	} else {
+		// Network Policy Azure can be used only with Network Plugin Azure CNI
+		if containerservice.NetworkPolicy(to.String(spec.NetworkPolicy)) == containerservice.NetworkPolicyAzure {
+			networkProfile.NetworkPolicy = containerservice.NetworkPolicyCalico
```
cmurphy, line 76: I don't think it's good practice to silently change a configured value behind the user's back like this. It should either be validated and an error returned to let the user correct it, or even just pass the invalid config to AKS and bubble up the error from there.

cmurphy (commented):
I'm not clear what this is fixing, is there an issue that goes with this?

pkg/aks/create.go
PennyScissors: @cmurphy This is fixing an issue with AKSv2 provisioning where the network configuration is not being set properly so the cluster gets deployed with no network policy enabled and kubenet as the default plugin in all cases. As for silently change a configured value, I agree that it should be validated or return an error. The UI has logic to make sure you can't select the Azure network policy without the Azure CNI already.

controller/aks-cluster-config-handler.go
```diff
@@ -403,9 +403,15 @@ func (h *Handler) validateConfig(config *aksv1.AKSClusterConfig) error {
 		return fmt.Errorf("at least one NodePool with mode System is required")
 	}
 
+	if config.Spec.NetworkPlugin != nil &&
+		to.String(config.Spec.NetworkPlugin) != string(containerservice.Kubenet) &&
+		to.String(config.Spec.NetworkPlugin) != string(containerservice.Azure) {
+		return fmt.Errorf("wrong network plugin value for [%s] cluster config", config.ClusterName)
```
rmweir, line 409:
```suggestion
		return fmt.Errorf("invalid network plugin value [%s] for [%s] cluster config", config.Spec.NetworkPlugin, config.ClusterName)
```

controller/aks-cluster-config-handler.go
```diff
@@ -403,9 +403,15 @@ func (h *Handler) validateConfig(config *aksv1.AKSClusterConfig) error {
...
+	}
+
 	if config.Spec.NetworkPolicy != nil &&
-		*config.Spec.NetworkPolicy != string(containerservice.NetworkPolicyAzure) &&
-		*config.Spec.NetworkPolicy != string(containerservice.NetworkPolicyCalico) {
+		to.String(config.Spec.NetworkPolicy) != string(containerservice.NetworkPolicyAzure) &&
+		to.String(config.Spec.NetworkPolicy) != string(containerservice.NetworkPolicyCalico) {
 		return fmt.Errorf("wrong network policy value for [%s] cluster config", config.ClusterName)
```
rmweir, line 415:
```suggestion
		return fmt.Errorf("invalid network policy value [%s] for [%s] cluster config", config.Spec.NetworkPolicy, config.ClusterName)
```

rmweir (changes requested):

pkg/aks/create.go
mjura: Comments included, thank you

controller/aks-cluster-config-handler.go
mjura: done

controller/aks-cluster-config-handler.go
mjura: done

controller/aks-cluster-config-handler.go
```diff
@@ -403,10 +403,20 @@ func (h *Handler) validateConfig(config *aksv1.AKSClusterConfig) error {
 		return fmt.Errorf("at least one NodePool with mode System is required")
 	}
 
+	if config.Spec.NetworkPlugin != nil &&
+		to.String(config.Spec.NetworkPlugin) != string(containerservice.Kubenet) &&
+		to.String(config.Spec.NetworkPlugin) != string(containerservice.Azure) {
+		return fmt.Errorf("invalid network plugin value [%s] for [%s] cluster config", to.String(config.Spec.NetworkPlugin), config.ClusterName)
+	}
+
 	if config.Spec.NetworkPolicy != nil &&
```
thedadams, line 412: nit: Can combine these to if-blocks.

thedadams (approved):

rmweir (dismissed):
LGTM

controller/aks-cluster-config-handler.go
```diff
@@ -403,10 +403,20 @@ func (h *Handler) validateConfig(config *aksv1.AKSClusterConfig) error {
...
-		*config.Spec.NetworkPolicy != string(containerservice.NetworkPolicyAzure) &&
-		*config.Spec.NetworkPolicy != string(containerservice.NetworkPolicyCalico) {
-		return fmt.Errorf("wrong network policy value for [%s] cluster config", config.ClusterName)
+		to.String(config.Spec.NetworkPolicy) != string(containerservice.NetworkPolicyAzure) &&
+		to.String(config.Spec.NetworkPolicy) != string(containerservice.NetworkPolicyCalico) {
+		return fmt.Errorf("invalid network policy value [%s] for [%s] cluster config", to.String(config.Spec.NetworkPolicy), config.ClusterName)
+	}
+
+	if !(to.String(config.Spec.NetworkPlugin) == string(containerservice.Azure) && to.String(config.Spec.NetworkPolicy) == string(containerservice.NetworkPolicyAzure)) {
+		return fmt.Errorf("azure network policy can be used only with Azure CNI network plugin for [%s],")
```
cmurphy, line 419: linter is failing here

```
controller/aks-cluster-config-handler.go:419:10: printf: Errorf format %s reads arg #1, but call has 0 args (govet)
464			return fmt.Errorf("azure network policy can be used only with Azure CNI network plugin for [%s],")
```

mjura: fixed

controller/aks-cluster-config-handler.go
mjura: done

cmurphy (approved):

rmweir (approved):

thedadams (approved):

PR #31, opened 2021-06-16
Check for enabled RBAC on the cluster

thedadams:
It is possible for a cluster to be created in AKS with RBAC disabled. In
order to warn a user that a particular cluster has RBAC disabled, the
status is fetched from AKS on import (and assumed true on create because
true is the default).

Issue:
https://github.com/rancher/rancher/issues/27670


cmurphy (commented):
Does rancher even work if rbac is disabled in the hosted cluster?

I think a better approach than adding a new parameter that doesn't have any functional use is to have the UI call /meta/aksClusters and look up the cluster that way:

https://github.com/rancher/rancher/blob/65e625a8561ab370c9645ca60ee1a3b2de1ce803/pkg/api/norman/customization/aks/listers.go#L225

cmurphy:
> I think a better approach than adding a new parameter that doesn't have any functional use is to have the UI call /meta/aksClusters and look up the cluster that way:

Which I guess is what https://github.com/rancher/rancher/pull/33209 is doing

mjura (approved):
LGTM

thedadams:
@cmurphy It does "work" in that you can import the cluster and things seem fine. I haven't poked around it too much.

We are waiting to hear back as to whether https://github.com/rancher/rancher/pull/33209 is enough (I hope that it is) or whether we need this. Ultimately, the new parameter on the GKE config object is so a parameter can be added to the cluster object, which then can be used by the UI to display a warning about RBAC being disabled.

rmweir (approved):
LGTM

cmurphy (approved):

controller/aks-cluster-config-handler.go
```diff
@@ -283,6 +283,12 @@ func (h *Handler) checkAndUpdate(config *aksv1.AKSClusterConfig) (*aksv1.AKSClus
 		return config, err
 	}
 
+	if config.Status.RBACEnabled == nil && result.EnableRBAC != nil {
+		config = config.DeepCopy()
+		config.Status.RBACEnabled = result.EnableRBAC
+		return h.aksCC.UpdateStatus(config)
```
cbron, line 289: >  In order to warn a user....

Meaning that this will just exist on status and the UI will be the one displaying the error message ? We'll need to make a UI issue (if Wes isn't already working on this one).

cbron (approved):

cbron:
Both could work right ? You'd want the list check for registering either way. Then its just whether the cluster page itself would use this or use the list call.

thedadams:
@cbron Yes, this enables the "continuous" message on the cluster view. The rancher/rancher PR integrates these changes, as well as including RBAC status when listing clusters for registering.

PR #33, opened 2021-06-17
Improve validation for Azure CNI

mjura:


controller/aks-cluster-config-handler.go
```diff
@@ -416,6 +416,24 @@ func (h *Handler) validateConfig(config *aksv1.AKSClusterConfig) error {
...
+		}
+		if config.Spec.NetworkDNSServiceIP == nil {
+			return fmt.Errorf(cannotBeNilErrorAzurePlugin, "dnsServiceIp", config.ClusterName)
+		}
+		if config.Spec.NetworkDockerBridgeCIDR == nil {
+			return fmt.Errorf(cannotBeNilErrorAzurePlugin, "dockerBridgeCidr", config.ClusterName)
+		}
+		if config.Spec.NetworkServiceCIDR == nil {
+			return fmt.Errorf(cannotBeNilErrorAzurePlugin, "serviceCidr", config.ClusterName)
+		}
```
rmweir, line 435: What if a cluster was registered and then the NetworkPlugin was changed? Would these errors be encountered?

cmurphy: I think it's safe because this will bail early if it's a registered cluster https://github.com/rancher/aks-operator/pull/33/files#diff-1aa54d4941ca80708078867e1b130aff086719689479dea38683fded2f3c7852R362-R364

But for GKE and EKS we require all top-level parameters are set, even if they're empty and not going to be used. Should we be doing that here as well, regardless of the network plugin?

mjura: After cluster was provisioned, NetworkPlugin can't be changed. This applies to the most hosted Kubernetes clusters.

rmweir (approved):
LGTM

thedadams (approved):

PR #34, opened 2021-06-22
Use same SLES image as Rancher

thedadams:
In addition to using the SLES image for the aks-operator image, the
dapper image is also switched to using SLES.
    
A side effect of this is that docker will no longer be installed in the
dapper image. Instead, a new pipeline step is added to build the image
on pull requests.

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -1,12 +1,21 @@
-FROM golang:1.16
+FROM registry.suse.com/suse/sle15:15.3
 
 ARG DAPPER_HOST_ARCH
-ENV ARCH $DAPPER_HOST_ARCH
+ENV ARCH=${DAPPER_HOST_ARCH}
 
-RUN apt update && \
-    apt install -y bash git gcc docker.io vim less file curl wget ca-certificates
+RUN zypper -n update && \
+    zypper -n install bash git binutils glibc-devel-static gcc vim less file tar gzip curl sed wget ca-certificates
+
+ENV GOLANG_ARCH_amd64=amd64 GOLANG_ARCH_arm=armv6l GOLANG_ARCH_arm64=arm64 GOLANG_ARCH=GOLANG_ARCH_${ARCH} \
+    GOPATH=/go PATH=/go/bin:/usr/local/go/bin:${PATH} SHELL=/bin/bash
+RUN curl -sLf https://storage.googleapis.com/golang/go1.16.4.linux-${ARCH}.tar.gz | tar -xzf - -C /usr/local/
+# workaround for https://bugzilla.suse.com/show_bug.cgi?id=1183043
+RUN if [ "${ARCH}" == "arm64" ]; then \
+        zypper -n install binutils-gold ; \
+    fi
+
 RUN if [ "${ARCH}" = "amd64" ]; then \
-        curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.30.0; \
+        curl -sL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s v1.40.1; \
     fi
 RUN curl -sL https://get.helm.sh/helm-v3.3.0-linux-${ARCH}.tar.gz | tar xvzf - -C /usr/local/bin --strip-components=1
 
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -1,8 +1,9 @@
-ARG UBI_IMAGE=registry.access.redhat.com/ubi8/ubi-minimal:latest
-FROM ${UBI_IMAGE}
-ENV KUBECONFIG /root/.kube/config
-RUN microdnf update -y && \
-    rm -rf /var/cache/yum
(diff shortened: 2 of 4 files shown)


mjura (dismissed):
Finally AWESOME!!!

cmurphy (approved):

rmweir (approved):
LGTM

package/Dockerfile
```diff
@@ -1,8 +1,10 @@
...
+FROM ${SLES_IMAGE}
+RUN zypper update -y && \
+    zypper -n clean -a && \
+    rm -rf /tmp/* /var/tmp/* /usr/share/doc/packages/*
+RUN useradd --uid 1007 aks-operator
+ENV KUBECONFIG /home/aks-operator/.kube/config
 COPY bin/aks-operator /usr/bin/
-USER 1001
+USER 1007
 ENTRYPOINT ["aks-operator"]
```
cbron, line 10: Why not move the actual Dockerfile ?

mjura (dismissed):

Dockerfile.dapper
```diff
@@ -1,20 +1,35 @@
...
+    DOCKER_URL_arm64=https://github.com/rancher/docker/releases/download/v1.10.3-ros1/docker-1.10.3_arm64 \
+    DOCKER_URL=DOCKER_URL_${ARCH}
+RUN curl -sLf ${!DOCKER_URL} -o /usr/bin/docker && chmod +x /usr/bin/docker
+
 ENV DAPPER_ENV REPO TAG DRONE_TAG CROSS
 ENV DAPPER_SOURCE /go/src/github.com/rancher/aks-operator/
 ENV DAPPER_OUTPUT ./bin ./dist
 ENV DAPPER_DOCKER_SOCKET true
-ENV DAPPER_RUN_ARGS "-v aks-operator-pkg:/go/pkg -v aks-operator-cache:/root/.cache"
+ENV DAPPER_RUN_ARGS "-v aks-operator-pkg:/go/pkg -v eks-operator-cache:/root/.cache"
```
cbron, line 32: revert

cbron (commented):
lgtm, just one typo

cbron:
Should wait to merge other PRs until this one is re-reviewed by others

Dockerfile.dapper
```diff
@@ -1,15 +1,30 @@
...
+# workaround for https://bugzilla.suse.com/show_bug.cgi?id=1183043
+RUN if [ "${ARCH}" == "arm64" ]; then \
+        zypper -n install binutils-gold ; \
+    fi
 
-RUN apt update && \
-    apt install -y bash git gcc docker.io vim less file curl wget ca-certificates
 RUN if [ "${ARCH}" = "amd64" ]; then \
-        curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.30.0; \
+        curl -sL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s v1.40.1; \
```
rmweir, line 18: Is there a "latest" equivalent we can use instead of 1.40.1?

cbron (dismissed):

Dockerfile.dapper
```diff
@@ -1,15 +1,30 @@
...
     fi
 RUN curl -sL https://get.helm.sh/helm-v3.3.0-linux-${ARCH}.tar.gz | tar xvzf - -C /usr/local/bin --strip-components=1
 
+ENV DOCKER_URL_amd64=https://get.docker.com/builds/Linux/x86_64/docker-1.10.3 \
+    DOCKER_URL_arm=https://github.com/rancher/docker/releases/download/v1.10.3-ros1/docker-1.10.3_arm \
+    DOCKER_URL_arm64=https://github.com/rancher/docker/releases/download/v1.10.3-ros1/docker-1.10.3_arm64 \
+    DOCKER_URL=DOCKER_URL_${ARCH}
```
rmweir, line 25: Is there a reason why we are using our own, outdated version of docker? Can we download docker with https://github.com/rancher/install-docker?

rmweir (changes requested):

Dockerfile.dapper
thedadams: No, it seems that a version is required to be specified: https://golangci-lint.run/usage/install/

Dockerfile.dapper
thedadams: I pulled this from rancher/rancher. I will switch it to use rancher/install-docker.

thedadams: This is why this doesn't use rancher/install-docker: you get an error
```
ERROR: Unsupported distribution 'sles'
```
Instead, I will use a more modern version of docker.

cmurphy (dismissed):

Dockerfile.dapper
dramich: Pin dependencies so builds don't end up with random failures.

rmweir (approved):
LGTM

thedadams:
PR updated to remove docker from the dapper image and use the docker plugin for building on PRs.

dramich (approved):

PR #36, opened 2021-07-09
Run c_rehash before launching the operator

thedadams:
The SLE image requires that c_rehash is run before additional trusted CAs
will be detected. Therefore, an entrypoint script is added to accomplish
this.

In addition, fsGroup is added so that c_rehash (which is run as the
aks-operator user) can read the cert that is mounted.

Issue:
https://github.com/rancher/rancher/issues/31846


mjura (approved):
LGTM

cmurphy (approved):

cbron (approved):

PR #37, opened 2021-07-12
Don't check node count in upstream when autoscaling enabled

mjura:


controller/aks-cluster-config-handler.go
```diff
@@ -704,14 +704,13 @@ func (h *Handler) updateUpstreamClusterState(ctx context.Context, secretsCache w
 			updateNodePool := false
 			upstreamNodePool, ok := upstreamNodePools[npName]
 			if ok {
-				if to.Bool(np.EnableAutoScaling) {
+				if to.Bool(np.EnableAutoScaling) && !to.Bool(upstreamNodePool.EnableAutoScaling) {
 					if to.Int32(np.Count) != to.Int32(upstreamNodePool.Count) {
 						return config, fmt.Errorf("can't update node count when autoscaling is enabled in node pool [%s] for cluster [%s]", to.String(np.Name), config.Spec.ClusterName)
 					}
```
rmweir, line 710: I think this should be checked in the block below as well.

rmweir (changes requested):

controller/aks-cluster-config-handler.go
rmweir, line 710: I think this should be checked in the block below as well. Another option is to refactor the logic so count doesn't need to be checked twice.

rmweir (changes requested):

controller/aks-cluster-config-handler.go
cmurphy: If I'm understanding the bug correctly (which I'm not sure I am) that would defeat the purpose, the point is to check that both EnableAutoScaling and Count aren't changing at the same time. But I agree that refactoring to avoid this quadrupal-nested conditional would be nice if possible.

controller/aks-cluster-config-handler.go
```diff
@@ -704,14 +704,13 @@ func (h *Handler) updateUpstreamClusterState(ctx context.Context, secretsCache w
...
-					if !to.Bool(upstreamNodePool.EnableAutoScaling) {
-						logrus.Infof("Enable autoscaling in node pool [%s] for cluster [%s]", to.String(np.Name), config.Spec.ClusterName)
-						updateNodePool = true
-					}
+					logrus.Infof("Enable autoscaling in node pool [%s] for cluster [%s]", to.String(np.Name), config.Spec.ClusterName)
+					updateNodePool = true
+				} else if to.Bool(np.EnableAutoScaling) && to.Bool(upstreamNodePool.EnableAutoScaling) {
```
cmurphy, line 713: Using `else if` here means that you can't change both `EnableAutoScaling` and `MinCount` or `MaxCount` at the same time, the change to `MinCount` would be ignored by the first case.

thedadams: It seems that `MinCount` and `MaxCount` can be changed regardless of the value of `EnableAutoScaling`. If this is correct, I think refactoring to make this section easier to comprehend would be valuable.

thedadams (changes requested):

controller/aks-cluster-config-handler.go
mjura: Current logic is OK. We have to recognize situation when:
- `EnableAutoScalling` is enabled on downstream and disabled on upstream, then `NodeCount` can't be changed, `MinCount` and `MaxCount` will be updated anyway
- `EnableAutoScalling` is enabled on downstream and upstream, then only  `MinCount` and `MaxCount` can be updated, `NodeCount` node count will be changed by AKS auto-scaler
- `EnableAutoScalling` is disabled in downstream and still enabled on upstream, then `MinCount` and `MaxCount` should be `nil`

This is Azure API logic

controller/aks-cluster-config-handler.go
mjura: This was already refactored and reflects all cases which we have to check with Azure API

thedadams (approved):

cmurphy (approved):

controller/aks-cluster-config-handler.go
rmweir: I thought that the point was, count cannot be changed if auto scaling is enabled.

thedadams:
If @rmweir and I understand this issue correctly, it is a bigger issue than just node count when `EnableAutoScaling` is `true`. The problem is that updates made in AKS can cause the cluster to be in a bad state in Rancher. One way is if `EnableAutoScaling` is `true` and AKS scales the node pool. In certain cases, `aks-operator` will see that the node count is not what is on the Spec and try to update it. This is not allowed.

Another scenario is when the user updates the control plane and node pool kubernetes version. AKS will update the control plane version first, followed by the node pool version. When the sync runs, it may sync the new control plane kubernetes version and the old node pool kubernetes version. When the update completes on AKS, then aks-operator will try to reconcile the state causing it to try to "update" the node pool kubernetes version which is an invalid update.

All of this is to say that this PR can be closed in favor of https://github.com/rancher/aks-operator/pull/42, which fixes all of these issues.

mjura:
I am closing this in favor of https://github.com/rancher/aks-operator/pull/42

PR #38, opened 2021-07-13
Don't attempt to update immutable properties

cmurphy:
Without this change, setting ServicePrincipalProfile on a cluster update
causes an error: "Changing property 'servicePrincipalProfile.clientId'
is not allowed", even though the clientID is not different. This change
checks whether the cluster is in an updating state before trying to set
the ServicePrincipalProfile on the cluster object.

Additionally, since DNSPrefix is nil for imported clusters, this value
was being changed to the cluster name, which might not be the true
value, and so was causing the operator to try to change it even though
it is immutable. This change makes it a required property, which is
already the case in UI, and ignores it if it is nil which will be the
case for imported clusters.

https://github.com/rancher/rancher/issues/33536

### DIFF
--- a/controller/aks-cluster-config-handler.go
+++ b/controller/aks-cluster-config-handler.go
@@ -237,7 +237,7 @@ func (h *Handler) createCluster(config *aksv1.AKSClusterConfig) (*aksv1.AKSClust
 		return config, err
 	}
 
-	err = aks.CreateOrUpdateCluster(ctx, credentials, resourceClusterClient, &config.Spec)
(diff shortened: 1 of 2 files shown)


cmurphy:
No UI change needed

thedadams (dismissed):

mjura (dismissed):
LGTM

cmurphy:
Fixed commit message

rmweir (approved):
LGTM

thedadams (approved):

PR #40, opened 2021-07-13
Use OsDiskType in request to create node pools

thedadams:
The OsDiskType was mistakenly leftout of the request to create node
pools, which caused the disk type to always default to Managed. Now the
setting should reflect what is in the object.

Issue:
https://github.com/rancher/rancher/issues/33603


mjura (approved):
LGTM

cmurphy (approved):

cbron (approved):

PR #42, opened 2021-07-14
Wait for non-Rancher initiated updates to complete

thedadams:
If a user updates the AKS cluster in AKS console, a transitory state may
be synced. Once Rancher sees this middle state, it will try to reconcile
back to this undesired state after the update completes in AKS.

With this change, aks-operator recognizes when an update was non
initiated by Rancher. It will then ignore the update until the AKS
cluster is in a "Succeeded" state.

Co-authored-by: Ricardo Weir <ricardo@rancher.com>

Issue:
https://github.com/rancher/rancher/issues/33446

### DIFF
--- a/controller/aks-cluster-config-handler.go
+++ b/controller/aks-cluster-config-handler.go
@@ -59,16 +59,20 @@ const (
 	// ClusterStatusDeleting The Deleting state indicates that cluster was removed, return code 4
 	ClusterStatusDeleting = "Deleting"
 
-	// NodePoolCreating The Creating state indicates that cluster was removed, return code 4
+	// NodePoolSucceeded The Succeeeded state indicates the node pool has been
+	// created and is fully usable, return code 0
+	NodePoolSucceeded = "Succeeded"
+
+	// NodePoolCreating The Creating state indicates that node pool is creating
 	NodePoolCreating = "Creating"
 
-	// NodePoolScaling The Scaling state indicates that cluster was removed, return code 4
+	// NodePoolScaling The Scaling state indicates that node pool is being scaled
 	NodePoolScaling = "Scaling"
 
-	// NodePoolDeleting The Deleting state indicates that cluster was removed, return code 4
+	// NodePoolDeleting The Deleting state indicates that node pool is being deleted
 	NodePoolDeleting = "Deleting"
 
-	// NodePoolUpgrading The Upgrading state indicates that cluster was upgraded
+	// NodePoolUpgrading The Upgrading state indicates that node pool is upgrading
 	NodePoolUpgrading = "Upgrading"
 )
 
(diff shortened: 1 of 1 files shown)


controller/aks-cluster-config-handler.go
```diff
@@ -294,6 +298,12 @@ func (h *Handler) checkAndUpdate(config *aksv1.AKSClusterConfig) (*aksv1.AKSClus
 		return config, fmt.Errorf("update failed for cluster [%s], status: %s", config.Spec.ClusterName, clusterState)
 	}
 	if clusterState == ClusterStatusInProgress || clusterState == ClusterStatusUpgrading {
+		// If the cluster is in an active state in Rancher but is updating in AKS, then an update was initiated in AKS console.
```
rmweir, line 301: Not necessarily from the console, just somewhere else.
```suggestion
		// If the cluster is in an active state in Rancher but is updating in AKS, then an update was initiated outside of rancher, such as through the AKS console.
```

rmweir (approved):
LGTM

controller/aks-cluster-config-handler.go
thedadams: Done

rmweir (approved):
LGTM

mjura (approved):
LGTM

controller/aks-cluster-config-handler.go
```diff
@@ -308,13 +318,11 @@ func (h *Handler) checkAndUpdate(config *aksv1.AKSClusterConfig) (*aksv1.AKSClus
 	for _, np := range *result.AgentPoolProfiles {
 		if status := to.String(np.ProvisioningState); status == NodePoolCreating ||
 			status == NodePoolScaling || status == NodePoolDeleting || status == NodePoolUpgrading {
-			if config.Status.Phase != aksConfigUpdatingPhase {
-				config = config.DeepCopy()
-				config.Status.Phase = aksConfigUpdatingPhase
-				config, err = h.aksCC.UpdateStatus(config)
```
cmurphy, line 314: There's nothing anymore that sets Phase to Updating if only node pools are changing, is that intended?

thedadams: I believe this will change the status when a node group is updated: https://github.com/thedadams/aks-operator/blob/cf75e6d76643265b31f4c791edd5959c1fb1190b/controller/aks-cluster-config-handler.go#L778

controller/aks-cluster-config-handler.go
```diff
@@ -708,9 +735,8 @@ func (h *Handler) updateUpstreamClusterState(ctx context.Context, secretsCache w
 			upstreamNodePool, ok := upstreamNodePools[npName]
 			if ok {
 				if to.Bool(np.EnableAutoScaling) {
-					if to.Int32(np.Count) != to.Int32(upstreamNodePool.Count) {
-						return config, fmt.Errorf("can't update node count when autoscaling is enabled in node pool [%s] for cluster [%s]", to.String(np.Name), config.Spec.ClusterName)
-					}
+					// Count can't be updated when EnableAutoScaling is true, so don't send anything.
+					np.Count = nil
```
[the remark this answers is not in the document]
cmurphy: Discussed offline, this is probably the best move. UI always sends a default value so it will have to be ignored here anyway in this case.

cmurphy (approved):

PR #43, opened 2021-07-16
Fix order of control plane and node pool upgrade

cmurphy:
This change does two things:

 1. Perform control plane upgrades first. This will make it possible
    to upgrade both the control plane and node pool kubernetes versions
    at the same time, because the controller will safely make the
    control plane changes first.
 2. Only make changes to the current upstream node pools in the
    cluster's AgentPoolProfiles list during the control plane update.
    Without this, if a node pool is added with a version different from
    the current upstream control plane version, AKS rejects it with this
    error:

      OrchestratorVersion must be all specified or all unspecified. If
      all specified, they must be stay unchanged or the same with
      control plane.

    Filter out the new ones, and then add them in the node pool update
    section after the control plane update is finished.

https://github.com/rancher/rancher/issues/33426

### DIFF
--- a/controller/aks-cluster-config-handler.go
+++ b/controller/aks-cluster-config-handler.go
@@ -718,6 +718,68 @@ func (h *Handler) updateUpstreamClusterState(ctx context.Context, secretsCache w
 		}
 	}
 
+	updateAksCluster := false
+	// check Kubernetes version for update
+	if config.Spec.KubernetesVersion != nil {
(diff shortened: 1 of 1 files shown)


cmurphy:
for 2.6.1

thedadams (approved):

rmweir (approved):
LGTM

cbron (approved):

PR #41, opened 2021-07-14
Update wrangler and k8s.io dependencies

cmurphy:
## Update wrangler for newer CRD generation

Latest wrangler generates version apiextensions.k8s.io/v1 CRDs which are
compatible with kubernetes 1.22. Required minor adjustments in the
codegen package.

---

## Update k8s.io dependencies to latest

---

https://github.com/rancher/rancher/issues/33622

PR #44, opened 2021-07-22
Lookup credential tenant ID if not provided

cmurphy:
* This change adds a dependency on the latest version of rancher/machine and therefore needs to pull in https://github.com/rancher/aks-operator/pull/41 to get the latest wrangler and k8s dependencies.

* Updates AKS dependencies to align with rancher/machine

* Use rancher/machine FindTenantID to look up the tenant ID if not found in the cloud credential secret

https://github.com/rancher/rancher/issues/33097


pkg/aks/client.go
```diff
@@ -133,5 +140,31 @@ func GetSecrets(secretsCache wranglerv1.SecretCache, spec *aksv1.AKSClusterConfi
...
+				return nil, err
+			}
+			secret.Annotations[tenantIDAnnotation] = cred.TenantID
+			secret.Annotations[tenantIDTimestampAnnotation] = time.Now().UTC().Format(time.RFC3339)
+			// Ignore errors when updating the secret object. If the secret cannot be updated
+			// (perhaps due to a conflict error), the tenant ID will be re-fetched on the next reconcile loop.
+			_, err := secretClient.Update(secret)
+			logrus.Debugf("encountered error while updating secret, ignoring: %v", err)
+		}
+	}
```
rmweir, line 167: Can this be put in a function that can be reused by rancher but with the last part being separated so the parent function can decide what to do with the error?

pkg/aks/client.go
```diff
@@ -133,5 +140,31 @@ func GetSecrets(secretsCache wranglerv1.SecretCache, spec *aksv1.AKSClusterConfi
...
+		if cred.TenantID == "" {
+			cred.TenantID, err = azureutil.FindTenantID(context.Background(), azure.PublicCloud, string(cred.SubscriptionID))
+			if err != nil {
+				return nil, err
+			}
+			secret.Annotations[tenantIDAnnotation] = cred.TenantID
+			secret.Annotations[tenantIDTimestampAnnotation] = time.Now().UTC().Format(time.RFC3339)
+			// Ignore errors when updating the secret object. If the secret cannot be updated
+			// (perhaps due to a conflict error), the tenant ID will be re-fetched on the next reconcile loop.
+			_, err := secretClient.Update(secret)
```
rmweir, line 164: The error should be nil checked and only printed out if it is not nil.

rmweir (changes requested):

pkg/aks/client.go
cmurphy: done

pkg/aks/client.go
cmurphy: fixed

pkg/codegen/main.go
```diff
@@ -53,11 +54,11 @@ func main() {
 		panic(err)
 	}
 
-	obj.ObjectMeta.Annotations = map[string]string{
+	obj.(*unstructured.Unstructured).SetAnnotations(map[string]string{
```
cbron, line 57: Is ObjectMeta gone ? We need to guarantee this annotation still gets set.

pkg/aks/client.go
```diff
@@ -133,5 +141,45 @@ func GetSecrets(secretsCache wranglerv1.SecretCache, spec *aksv1.AKSClusterConfi
...
+		parsedTime, err := time.Parse(time.RFC3339, timestamp)
+		if err != nil {
+			return "", err
+		}
+		if parsedTime.Add(tenantIDTimeout).After(time.Now().UTC()) {
+			logrus.Infof("using cached tenant ID")
+			return tenantAnno, nil
+		}
+	}
+	logrus.Infof("looking up tenant ID")
```
cbron, line 167: Should make these log messages more specific, include secret name, but not include any other cred info.

cbron (commented):
looks good, just minor things

pkg/aks/client.go
cmurphy: These were debug logs that I accidentally left in

pkg/codegen/main.go
cmurphy: [This change](https://github.com/rancher/wrangler/commit/c5ae8c7a7f241b4d2e766e965ccc7b6f85515f62#diff-099f7d83f8d29c3f530aa27406ae07b149faf56d91c1967c17eee2052d08197aR193) changed `obj` from `CustomResourceDefinition` to `runtime.Object`, which doesn't have an `ObjectMeta` property. What this is doing is casting the object to an `Unstructured` which we can call `SetAnnotations` on. The end result is the same, the annotation is set on the object.

thedadams (commented):
One minor comment right now.

go.mod
```diff
@@ -2,22 +2,22 @@ module github.com/rancher/aks-operator
...
 	github.com/sirupsen/logrus v1.4.2
 	github.com/stretchr/testify v1.7.0
-	golang.org/x/crypto v0.0.0-20210513164829-c07d793c2f9a // indirect
-	k8s.io/api v0.18.8
-	k8s.io/apimachinery v0.18.8
-	k8s.io/client-go v0.18.8
+	golang.org/x/crypto v0.0.0-20210711020723-a769d52b0f97 // indirect
+	k8s.io/api v0.21.2
+	k8s.io/apimachinery v0.21.2
+	k8s.io/client-go v0.21.0
```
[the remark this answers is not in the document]
cmurphy: done

thedadams (approved):

cbron (dismissed):
lgtm, ricky will review soon

pkg/aks/client.go
```diff
@@ -133,5 +142,39 @@ func GetSecrets(secretsCache wranglerv1.SecretCache, spec *aksv1.AKSClusterConfi
...
 	cred.BaseURL = spec.BaseURL
 
+	if cred.TenantID == "" {
+		cred.TenantID, err = GetCachedTenantID(cred.SubscriptionID, secret)
+		if err != nil {
+			return nil, err
+		}
+		secret.Annotations[tenantIDAnnotation] = cred.TenantID
+		secret.Annotations[tenantIDTimestampAnnotation] = time.Now().UTC().Format(time.RFC3339)
+		_, err = secretClient.Update(secret)
```
rmweir, line 152: Should only update these if timestamp was expired and value was not from annotation.

rmweir (changes requested):

cmurphy:
I am putting the secret update back inside GetCachedTenantID, that's the only place it makes sense to put it since that's the part that knows whether the tenant ID was re-fetched. I'm also handling the potential conflict error there. The two callers we have always want this error ignored, so moving the error handling outside the function just duplicates work.

cmurphy:
And added debug log

pkg/aks/client.go
```diff
@@ -133,5 +142,44 @@ func GetSecrets(secretsCache wranglerv1.SecretCache, spec *aksv1.AKSClusterConfi
...
+			return "", err
+		}
+		if parsedTime.Add(tenantIDTimeout).After(time.Now().UTC()) {
+			return tenantAnno, nil
+		}
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	logrus.Debugf("retrieving tenant ID from Azure public cloud")
+	tenantID, err := azureutil.FindTenantID(ctx, azure.PublicCloud, subscriptionID)
```
thedadams, line 174: Should this error be checked?

pkg/aks/client.go
```diff
@@ -133,5 +142,44 @@ func GetSecrets(secretsCache wranglerv1.SecretCache, spec *aksv1.AKSClusterConfi
...
+			return nil, err
+		}
+	}
+
 	return &cred, nil
 }
+
+type secretClient interface {
+	Update(*v1.Secret) (*v1.Secret, error)
+}
```
thedadams, line 157: Is this type necessary?

thedadams (changes requested):

pkg/aks/client.go
cmurphy: Oops, fixed

pkg/aks/client.go
cmurphy: I'm adding it to try to make this usable both here and by rancher. Here we use `SecretClient` from wrangler, there we have `SecretInterface` from norman, and trying to cast between the two or add new dependencies to one or the other seemed like a headache.

thedadams: Ah, that makes sense!

thedadams (approved):

rmweir (approved):
LGTM

mjura (approved):
LGTM

PR #45, opened 2021-08-18
fixed addon profiles overwriting each other

paynejacob:
https://github.com/rancher/rancher/issues/34244

# PROBLEM
If you enable container monitoring, httpApplicationRouting would never turn on.

# SOLUTION
The addonProfiles were being overwritten instead of appended too.  This caused the operator to continue updating the cluster infinitely.  I fixed this logic so that an combination of monitoring and http application routing can be configured.  I also fixed the operator attempting to update the cluster when it was already in an updating state.

### DIFF
--- a/controller/aks-cluster-config-handler.go
+++ b/controller/aks-cluster-config-handler.go
@@ -50,8 +50,8 @@ const (
 	// node software, return code 3
 	ClusterStatusInProgress = "InProgress"
 
-	// ClusterStatusUpgrading The Upgrading state indicates the cluster is updating
-	ClusterStatusUpgrading = "Upgrading"
+	// ClusterStatusUpdating The Updating state indicates the cluster is updating
+	ClusterStatusUpdating = "Updating"
 
 	// ClusterStatusCanceled The Canceled state indicates that create or update was canceled, return code 2
 	ClusterStatusCanceled = "Canceled"
@@ -297,7 +297,7 @@ func (h *Handler) checkAndUpdate(config *aksv1.AKSClusterConfig) (*aksv1.AKSClus
 	if clusterState == ClusterStatusFailed {
 		return config, fmt.Errorf("update failed for cluster [%s], status: %s", config.Spec.ClusterName, clusterState)
 	}
-	if clusterState == ClusterStatusInProgress || clusterState == ClusterStatusUpgrading {
+	if clusterState == ClusterStatusInProgress || clusterState == ClusterStatusUpdating {
 		// If the cluster is in an active state in Rancher but is updating in AKS, then an update was initiated outside of Rancher,
 		// such as in AKS console. In this case, this is a no-op and the reconciliation will happen after syncing.
 		if config.Status.Phase == aksConfigActivePhase {
--- a/pkg/aks/create.go
+++ b/pkg/aks/create.go
@@ -110,21 +110,17 @@ func CreateOrUpdateCluster(ctx context.Context, cred *Credentials, clusterClient
 		}
 	}
 
-	var addonProfiles map[string]*containerservice.ManagedClusterAddonProfile
+	addonProfiles := make(map[string]*containerservice.ManagedClusterAddonProfile, 0)
 
 	if hasHTTPApplicationRoutingSupport(spec) {
-		addonProfiles = map[string]*containerservice.ManagedClusterAddonProfile{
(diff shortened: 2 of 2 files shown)


controller/aks-cluster-config-handler.go
```diff
@@ -50,8 +50,8 @@ const (
 	// node software, return code 3
 	ClusterStatusInProgress = "InProgress"
 
-	// ClusterStatusUpgrading The Upgrading state indicates the cluster is updating
-	ClusterStatusUpgrading = "Upgrading"
+	// ClusterStatusUpdating The Upgrading state indicates the cluster is updating
```
snasovich, line 53:
```suggestion
	// ClusterStatusUpdating The Updating state indicates the cluster is updating
```

pkg/aks/create.go
```diff
@@ -110,21 +110,17 @@ func CreateOrUpdateCluster(ctx context.Context, cred *Credentials, clusterClient
 		}
 	}
 
-	var addonProfiles map[string]*containerservice.ManagedClusterAddonProfile
+	addonProfiles := map[string]*containerservice.ManagedClusterAddonProfile{}
```
snasovich, line 113: `make` is preferrable in this caseper our standards (which are basically https://github.com/uber-go/guide/blob/master/style.md#initializing-maps)

pkg/aks/create.go
```diff
@@ -110,21 +110,17 @@ func CreateOrUpdateCluster(ctx context.Context, cred *Credentials, clusterClient
...
 		}
 	}
 
 	if to.Bool(spec.Monitoring) {
-		addonProfiles = map[string]*containerservice.ManagedClusterAddonProfile{
-			"omsAgent": {
-				Enabled: spec.Monitoring,
-			},
+		addonProfiles["omsAgent"] = &containerservice.ManagedClusterAddonProfile{
+			Enabled: spec.HTTPApplicationRouting,
```
snasovich, line 123: Should it be `spec.Monitoring` here?

snasovich (changes requested):

cmurphy (approved):

snasovich (approved):

paynejacob:
This pr will also resolve https://github.com/rancher/rancher/issues/34161

controller/aks-cluster-config-handler.go
```diff
@@ -50,8 +50,8 @@ const (
 	// node software, return code 3
 	ClusterStatusInProgress = "InProgress"
 
-	// ClusterStatusUpgrading The Upgrading state indicates the cluster is updating
-	ClusterStatusUpgrading = "Upgrading"
+	// ClusterStatusUpdating The Updating state indicates the cluster is updating
+	ClusterStatusUpdating = "Updating"
```
cbron, line 54: Looks like this is purely grammatical, no functional change.

pkg/aks/create.go
```diff
@@ -110,21 +110,17 @@ func CreateOrUpdateCluster(ctx context.Context, cred *Credentials, clusterClient
 		}
 	}
 
-	var addonProfiles map[string]*containerservice.ManagedClusterAddonProfile
+	addonProfiles := make(map[string]*containerservice.ManagedClusterAddonProfile, 0)
```
cbron, line 113: I was thinking this would cause a problem limiting at zero, but it doesn't.

cbron (approved):

pkg/aks/create.go
paynejacob: For maps it does not restrict the size. If you know how many keys you will have it saves a ton of time with allocations.

controller/aks-cluster-config-handler.go
paynejacob: This does change the behavior in that the operator actually waits to apply changes now.  Before it would log constant error messages because you tried to patch a cluster in an updating state.

pkg/aks/create.go
snasovich: Yeah, it's capacity hint - not size. I believe you can omit this too and it would be the same as passing 0.

controller/aks-cluster-config-handler.go
cbron: Talked offline. There aren't good docs around this, but was proved out with some testing.

cbron (approved):

PR #48, opened 2021-11-01
Allow AKS clusters with same name from different subscriptions

polivbr:
Addresses https://github.com/rancher/rancher/issues/35312


polivbr:
Also requires: https://github.com/rancher/rancher/pull/35335

PR #54, opened 2022-02-23
Add checks for misconfigured oms agent

annablender:
Issue: https://github.com/rancher/rancher/issues/36425 

 
## Problem
 

 
The customer had an Azure setup where they had log analytics enabled on an AKS cluster, but either didn't have the OMS agent configured or it was misconfigured. This caused a panic in Rancher. I reproduced the issue internally, and fixed it by adding checks for a nil or empty OMS agent profile configuration.

## Testing
 

1. Added test code to a local copy of `rancher/aks-operator` simulating a nil or misconfigured OMS agent profile in an Azure portal setup
2. Ran standalone `ask-operator`
3. Ran an instance of Rancher v2.6-head and scaled the ask-operator deployment count to 0
4. Created an AKS cluster through Rancher
5. Checked that an invalid oms agent configuration error is seen in the UI

image

image

### DIFF
--- a/controller/aks-cluster-config-handler.go
+++ b/controller/aks-cluster-config-handler.go
@@ -665,6 +665,10 @@ func BuildUpstreamClusterState(ctx context.Context, secretsCache wranglerv1.Secr
 	// set addon monitoring profile
 	if addonProfile["omsAgent"] != nil {
 		upstreamSpec.Monitoring = addonProfile["omsAgent"].Enabled
+
+		if len(addonProfile["omsAgent"].Config) == 0 {
+			return nil, fmt.Errorf("cannot set OMS Agent configuration retrieved from Azure")
+		}
 		logAnalyticsWorkspaceResourceID := addonProfile["omsAgent"].Config["logAnalyticsWorkspaceResourceID"]
 
 		logAnalyticsWorkspaceGroup := matchWorkspaceGroup.FindStringSubmatch(to.String(logAnalyticsWorkspaceResourceID))[1]


controller/aks-cluster-config-handler.go
```diff
@@ -665,6 +665,10 @@ func BuildUpstreamClusterState(ctx context.Context, secretsCache wranglerv1.Secr
 	// set addon monitoring profile
 	if addonProfile["omsAgent"] != nil {
 		upstreamSpec.Monitoring = addonProfile["omsAgent"].Enabled
+
+		if addonProfile["omsAgent"].Config == nil || len(addonProfile["omsAgent"].Config) == 0 {
```
thedadams, line 669: if `.Config` is `nil`, then it will have length of 0.
```suggestion
		if len(addonProfile["omsAgent"].Config) == 0 {
```

controller/aks-cluster-config-handler.go
```diff
@@ -665,6 +665,10 @@ func BuildUpstreamClusterState(ctx context.Context, secretsCache wranglerv1.Secr
...
+			return nil, fmt.Errorf("cannot set OMS Agent configuration retrieved from Azure")
```
thedadams, line 670: Can we have a more descriptive error here? Why are we getting a misconfigured response from Azure?

thedadams (changes requested):

controller/aks-cluster-config-handler.go
annablender: Originally, I had 'cannot set invalid OMS agent configuration on cluster <cluster name>' but @dnoland1 requested the message above because we don't know if it's an invalid profile configuration. It's just a configuration that's nil or that we don't expect.

I'm not sure why the customer got a misconfigured OMS profile from Azure. We don't have enough info from them to know, we just know that whatever profile got returned didn't have .Config set.

controller/aks-cluster-config-handler.go
annablender: Ah - Go read operations work even when data structs are nil. This is different from Java, thanks for pointing that out

thedadams (approved):

jiaqiluo (approved):

PR #58, opened 2022-05-11
Add default Azure tags to config spec to fix sync loop

annablender:
Issue: https://github.com/rancher/rancher/issues/37608 

 
## Problem
 

When a private AKS cluster is provisioned, `aks-operator` tries to eternally remove auto-generated tags from the upstream spec which prevents the cluster from becoming active.
 
## Solution
 

The solution here is for rancher to move on if a tags update returns an unexpected response from Azure. In other words, if Azure doesn't return the requested state for the tags string map and also an HTTP OK, then the `aks-operator` will move on instead of enqueuing a tags update. Any tags automatically added back to the cluster resource by Azure will be added back to rancher on the next sync.

 
## Testing
 

* Provision an AKS cluster using the Rancher UI (it can be non-private. The bug is seen regardless)
* Verify that the default Azure tags were synced to the Rancher config object by checking that the cluster is active and the tags appear in the cluster yaml.
* Remove all tags
* Check that the default ones from Azure were synced back to rancher

image

image

### DIFF
--- a/controller/aks-cluster-config-handler.go
+++ b/controller/aks-cluster-config-handler.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"net/http"
 	"reflect"
 	"regexp"
+	"strings"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2020-11-01/containerservice"
@@ -21,6 +23,7 @@ import (
 	v15 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/util/retry"
 )
 
 const (
@@ -693,7 +696,7 @@ func BuildUpstreamClusterState(ctx context.Context, secretsCache wranglerv1.Secr
 }
 
 // updateUpstreamClusterState compares the upstream spec with the config spec, then updates the upstream AKS cluster to
-// match the config spec. Function returns after a update is finished.
+// match the config spec. Function returns after an update is finished.
 func (h *Handler) updateUpstreamClusterState(ctx context.Context, secretsCache wranglerv1.SecretCache,
 	secretClient wranglerv1.SecretClient,
 	config *aksv1.AKSClusterConfig, upstreamSpec *aksv1.AKSClusterConfigSpec) (*aksv1.AKSClusterConfig, error) {
@@ -709,16 +712,33 @@ func (h *Handler) updateUpstreamClusterState(ctx context.Context, secretsCache w
 
 	// check tags for update
 	if config.Spec.Tags != nil {
+
 		if !reflect.DeepEqual(config.Spec.Tags, upstreamSpec.Tags) {
 			logrus.Infof("Updating tags for cluster [%s]", config.Spec.ClusterName)
 			tags := containerservice.TagsObject{
 				Tags: *to.StringMapPtr(config.Spec.Tags),
 			}
-			_, err = resourceClusterClient.UpdateTags(ctx, config.Spec.ResourceGroup, config.Spec.ClusterName, tags)
+			response, err := resourceClusterClient.UpdateTags(ctx, config.Spec.ResourceGroup, config.Spec.ClusterName, tags)
 			if err != nil {
 				return config, err
 			}
-			return h.enqueueUpdate(config)
+
+			// Azure may have a policy that automatically adds upstream default tags to a cluster resource. We don't
+			// have a good way to detect that policy. We handle this case by checking if Azure returns an unexpected
+			// state for the tags and if so, log the response and move on. Any upstream tags regenerated on the cluster
+			// by Azure will be synced back to rancher.
+			upstreamTags := containerservice.TagsObject{}
+			err = retry.OnError(retry.DefaultBackoff, func(err error) bool { return strings.HasSuffix(err.Error(), "asynchronous operation has not completed") }, func() error {
+				managedCluster, err := response.Result(*resourceClusterClient)
+				upstreamTags.Tags = managedCluster.Tags
+				return err
+			})
+
+			if !reflect.DeepEqual(tags, upstreamTags) && response.Response().StatusCode == http.StatusOK {
+				logrus.Infof("Tags were not updated as expected for cluster [%s], expected %s, actual %s, moving on", config.Spec.ClusterName, to.StringMap(tags.Tags), to.StringMap(upstreamTags.Tags))
+			} else {
+				return h.enqueueUpdate(config)
+			}
 		}
 	}
 


annablender:
@thedadams There's no `release/v2.6` branch in the aks-operator repo. What do you mean?

thedadams:
My apologies. I didn't realize this was an `aks-operator` and not a `rancher/rancher` change. Please forgive my sloppiness.

jiaqiluo (approved):

thedadams (changes requested):
This fix relies on the syncing feature. I believe the following two scenarios will still cause problems:
Firstly:
1. Create an AKS cluster without specifying tags.
2. While the cluster is creating, change anything in the AKS cluster configuration (like specify a tag).
3. The cluster will get stuck as it previously did.

Secondly:
1. Create an AKS cluster with or without specifying tags.
2. After the cluster is active and before the first sync runs (so immediately after the cluster is active) change something in the AKS cluster configuration.
3. The cluster will get stuck as it previously did.

I think the solution we come up with should handle both of the above situations.

controller/aks-cluster-config-handler.go
```diff
@@ -707,21 +709,36 @@ func (h *Handler) updateUpstreamClusterState(ctx context.Context, secretsCache w
...
 				return config, err
 			}
+			// Azure will regenerate default upstream tags and add them to the cluster resource if a user tries to
+			// remove them all in one request, but the Azure job that removes tags doesn't return an error in that
+			// case. So, we check whether Azure returns an unexpected state for the tags and if so, log the response
+			// and move on. Any upstream tags that were regenerated and put back on the cluster by Azure will be synced
+			// back to rancher.
+			if !reflect.DeepEqual(tags, upstreamSpec.Tags) && response.Status() == "Succeeded" {
+				logrus.Infof("Azure API response was not expected. Completing tag update for cluster [%s]", config.Spec.ClusterName)
+				UnexpectedTagUpdateResponseInAzure = true
```
thedadams, line 732: Why not just move forward here instead of re-enqueuing?

controller/aks-cluster-config-handler.go
```diff
@@ -707,21 +709,36 @@ func (h *Handler) updateUpstreamClusterState(ctx context.Context, secretsCache w
...
+			}
 			return h.enqueueUpdate(config)
 		}
 	}
 
+	if UnexpectedTagUpdateResponseInAzure {
+		UnexpectedTagUpdateResponseInAzure = false
+	}
```
thedadams, line 740: If a user has more than one AKS cluster, then this global variable gets used for all of them. That means that if one cluster has an issue updating tags, then another cluster may have its tag update inadvertently skipped. Which would cause issues for the cluster having trouble updating its tags.

controller/aks-cluster-config-handler.go
```diff
@@ -707,21 +709,36 @@ func (h *Handler) updateUpstreamClusterState(ctx context.Context, secretsCache w
...
+			response, err := resourceClusterClient.UpdateTags(ctx, config.Spec.ResourceGroup, config.Spec.ClusterName, tags)
 			if err != nil {
 				return config, err
 			}
+			// Azure will regenerate default upstream tags and add them to the cluster resource if a user tries to
+			// remove them all in one request, but the Azure job that removes tags doesn't return an error in that
+			// case. So, we check whether Azure returns an unexpected state for the tags and if so, log the response
+			// and move on. Any upstream tags that were regenerated and put back on the cluster by Azure will be synced
+			// back to rancher.
+			if !reflect.DeepEqual(tags, upstreamSpec.Tags) && response.Status() == "Succeeded" {
```
thedadams, line 730: Updates to objects in Azure are done asynchronously. It seems that the tags call is very quick, but is it possible to get to this line and have `response.Status() == ""`?

thedadams (changes requested):

controller/aks-cluster-config-handler.go
annablender: No I don't think so. Rancher waits for the tags call to return with a valid API response before moving on from the `UpdateTags` line.

controller/aks-cluster-config-handler.go
annablender: Yeahh, this is the tricky part. So in the case that the user removes all the upstream default tags, Azure will respond with the list of autogenerated ones. In order for the tags to sync back to rancher as we discussed, we still have to enqueue the update.

controller/aks-cluster-config-handler.go
annablender: Good thought! Testing..

controller/aks-cluster-config-handler.go
thedadams: The response is a `Future` which means it doesn't wait for the action to complete. The API response is returned, but the `.Status()` call is on the `Future`.

annablender: I didn't know that, thank you. Unsure if `response.Status()` could return "", but I updated the func to check the latest HTTP response code of the async call. That will most likely return a valid response code instead of possibly returning an empty string.

This is straight from the Azure SDK

```
// returns the cached HTTP response after a call to pollForStatus(), can be nil
latestResponse() *http.Response
```

controller/aks-cluster-config-handler.go
annablender: You're correct, I mixed up what I was seeing in the ui with what was happening on the backend and thought re-enqueuing was needed but that is not the case. Tags added back from Azure will get set on the rancher obj on the next sync. I updated the code

controller/aks-cluster-config-handler.go
```diff
@@ -709,16 +709,25 @@ func (h *Handler) updateUpstreamClusterState(ctx context.Context, secretsCache w
...
+			response, err := resourceClusterClient.UpdateTags(ctx, config.Spec.ResourceGroup, config.Spec.ClusterName, tags)
 			if err != nil {
 				return config, err
 			}
-			return h.enqueueUpdate(config)
+			// Azure may have a policy that automatically adds upstream default tags to a cluster resource. We don't
+			// have a good way to detect that policy. We handle this case by checking if Azure returns an unexpected
+			// state for the tags and if so, log the response and move on. Any upstream tags regenerated on the cluster
+			// by Azure will be synced back to rancher.
+			if !reflect.DeepEqual(tags, upstreamSpec.Tags) && response.Response().StatusCode == 200 {
```
thedadams, line 726: This if-check will always be there are updates to the tags. The `upstreamSpec` object doesn't get automatically updated when the API call is made.

Is there a way to get the tags from the `response`?

controller/aks-cluster-config-handler.go
```diff
@@ -709,16 +709,25 @@ func (h *Handler) updateUpstreamClusterState(ctx context.Context, secretsCache w
...
+				logrus.Infof("Azure API response was not expected. Completing tag update for cluster [%s]", config.Spec.ClusterName)
```
thedadams, line 727: Let's add a debug message here that prints out the tags on the config and the tags in the response so that turning on debug will allow a user to figure out which tags are causing this.

thedadams (changes requested):

controller/aks-cluster-config-handler.go
annablender: It doesn't? After the API call is made I always see the new set of tags with the update listed on the upstreamSpec object. If you mean I can't count on that result, thennn.. yes, I believe so

thedadams: That was probably when you were looping back through the second time in the previous implementation.

controller/aks-cluster-config-handler.go
```diff
@@ -870,3 +886,38 @@ func (h *Handler) updateUpstreamClusterState(ctx context.Context, secretsCache w
 	logrus.Infof("Configuration for cluster [%s] was verified", config.Spec.ClusterName)
 	return config, err
 }
+
+// GetDefaultUpstreamTags gets the tags returned from the response returned by an update tags request sent to the Azure
+// API. Tags are returned in a TagsObject struct.
+func GetDefaultUpstreamTags(response containerservice.ManagedClustersUpdateTagsFuture) (containerservice.TagsObject, error) {
```
thedadams, line 892: The word `Default` here is not correct. This is just getting the tags on the cluster after the request. There doesn't seem to be any concept of "default" in that scenario.

controller/aks-cluster-config-handler.go
```diff
@@ -870,3 +886,38 @@ func (h *Handler) updateUpstreamClusterState(ctx context.Context, secretsCache w
...
+		// Add the tag
+		if foundTags  {
+			tagsMap[arr[i]] = arr[i + 2]
+			i = i + 3
+		}
+		// Done with adding tags
+		if foundTags && strings.Contains(arr[i], "}") {
+			break
+		}
+	}
```
thedadams, line 918: Is the response body here JSON? If so, we shouldn't be parsing it manually. Instead, `json.Unmarshal` should be used with a proper type to avoid errors.

There is also a `Result` method on the response object that returns a `ManagedCluster` object that has presumably been parsed with the data you are looking for here.

controller/aks-cluster-config-handler.go
```diff
@@ -709,16 +711,30 @@ func (h *Handler) updateUpstreamClusterState(ctx context.Context, secretsCache w
...
 			if err != nil {
 				return config, err
 			}
-			return h.enqueueUpdate(config)
+			// Azure may have a policy that automatically adds upstream default tags to a cluster resource. We don't
+			// have a good way to detect that policy. We handle this case by checking if Azure returns an unexpected
+			// state for the tags and if so, log the response and move on. Any upstream tags regenerated on the cluster
+			// by Azure will be synced back to rancher.
+			upstreamTags, err := GetDefaultUpstreamTags(response)
+
```
thedadams, line 729: Remove line between getting the error and checking the error.

controller/aks-cluster-config-handler.go
```diff
@@ -709,16 +711,30 @@ func (h *Handler) updateUpstreamClusterState(ctx context.Context, secretsCache w
...
+			if err != nil || (!reflect.DeepEqual(tags, upstreamTags) && response.Response().StatusCode == 200) {
```
thedadams, line 730:
```suggestion
			if err != nil || (!reflect.DeepEqual(tags, upstreamTags) && response.Response().StatusCode == http.StatusOK) {
```

controller/aks-cluster-config-handler.go
```diff
@@ -709,16 +711,30 @@ func (h *Handler) updateUpstreamClusterState(ctx context.Context, secretsCache w
...
+				logrus.Infof("Azure API response was not expected. Tags were requested to be updated to [%1s]. " +
+					"Azure returned the updated tags as [%2s]. Completing tag update for cluster [%3s]", to.StringMap(tags.Tags),
```
thedadams, line 732:
```suggestion
				logrus.Infof("Azure API response was not expected. Tags were requested to be updated to [%1s]. Azure returned the updated tags as [%2s]. Completing tag update for cluster [%3s]", to.StringMap(tags.Tags),
```

thedadams: Also, things like `%1s` means a string formatted with width of `1`. I don't think that is what you want here.

thedadams (changes requested):

controller/aks-cluster-config-handler.go
annablender: Yes, it is. Ah, I was looking into doing a `json.Unmarshal` into a custom struct and that option seemed unreasonable with how many fields there were in the response. Missed that the result had a `ManagedCluster` struct with the updated tags. Lesson learned there for this operator

controller/aks-cluster-config-handler.go
```diff
@@ -709,16 +712,32 @@ func (h *Handler) updateUpstreamClusterState(ctx context.Context, secretsCache w
...
 				return config, err
 			}
-			return h.enqueueUpdate(config)
+
+			// Azure may have a policy that automatically adds upstream default tags to a cluster resource. We don't
+			// have a good way to detect that policy. We handle this case by checking if Azure returns an unexpected
+			// state for the tags and if so, log the response and move on. Any upstream tags regenerated on the cluster
+			// by Azure will be synced back to rancher.
+			managedCluster := containerservice.ManagedCluster{}
+			err = retry.OnError(retry.DefaultBackoff, func(err error) bool { return !apierror.IsNotFound(err) }, func() error {
```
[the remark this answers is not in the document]
thedadams: I made a typo on what should be checked here. It shouldn't be `operator` but rather `operation`.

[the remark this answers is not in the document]
annablender: Whoops, gotcha

controller/aks-cluster-config-handler.go
```diff
@@ -709,16 +712,32 @@ func (h *Handler) updateUpstreamClusterState(ctx context.Context, secretsCache w
...
+			// state for the tags and if so, log the response and move on. Any upstream tags regenerated on the cluster
+			// by Azure will be synced back to rancher.
+			managedCluster := containerservice.ManagedCluster{}
+			err = retry.OnError(retry.DefaultBackoff, func(err error) bool { return strings.HasSuffix(err.Error(), "asynchronous operation has not completed") }, func() error {
+				managedCluster, err = response.Result(*resourceClusterClient)
+				return err
+			})
+
+			if !reflect.DeepEqual(tags, managedCluster.Tags) && response.Response().StatusCode == http.StatusOK {
+				logrus.Infof("Tags were not updated as expected for cluster %s, expected %s, actual %s, moving on", to.StringMap(tags.Tags),to.StringMap(managedCluster.Tags), config.Spec.ClusterName)
```
thedadams, line 737: nit, for `go fmt`:
```suggestion
				logrus.Infof("Tags were not updated as expected for cluster %s, expected %s, actual %s, moving on", to.StringMap(tags.Tags), to.StringMap(managedCluster.Tags), config.Spec.ClusterName)
```

controller/aks-cluster-config-handler.go
```diff
@@ -709,16 +712,32 @@ func (h *Handler) updateUpstreamClusterState(ctx context.Context, secretsCache w
...
+			// have a good way to detect that policy. We handle this case by checking if Azure returns an unexpected
+			// state for the tags and if so, log the response and move on. Any upstream tags regenerated on the cluster
+			// by Azure will be synced back to rancher.
+			managedCluster := containerservice.ManagedCluster{}
+			err = retry.OnError(retry.DefaultBackoff, func(err error) bool { return strings.HasSuffix(err.Error(), "asynchronous operation has not completed") }, func() error {
+				managedCluster, err = response.Result(*resourceClusterClient)
+				return err
+			})
+
+			if !reflect.DeepEqual(tags, managedCluster.Tags) && response.Response().StatusCode == http.StatusOK {
```
thedadams, line 736: The first part of this if (`!reflect.DeepEqual(tags, managedCluster.Tags)`) will always be `true` because `tags` and `managedCluster.Tags` have different types.

thedadams (changes requested):

controller/aks-cluster-config-handler.go
jiaqiluo, line 737: change the order of arguments
```suggestion
				logrus.Infof("Tags were not updated as expected for cluster %s, expected %s, actual %s, moving on", config.Spec.ClusterName, to.StringMap(tags.Tags),to.StringMap(managedCluster.Tags))
```

jiaqiluo (changes requested):

thedadams (approved):

PR #59, opened 2022-06-08
Fix field overwrite for cluster creation/update actions in AKS

annablender:
Issue: https://github.com/rancher/rancher/issues/37681  

 
## Problem
 

When creating or updating an AKS cluster, the aks operator does not check to see if a `managedCluster` object already exists and creates a new `managedCluster` object with only fields it knows about. This results in some fields associated with an existing cluster in Azure being overwritten on the `CreateOrUpdate` API call in `create.go`. When the aks operator does a sync and builds a config based on the state of the upstream AKS cluster, there is also a bug where authorized IP ranges configured in Azure will get overwritten with a nil value because rancher doesn't dereference the *[]string properly. Thus, the authorized IP range won't show up in the UI.
 
## Solution
 

I've refactored `create.go` in the aks operator to separate out the create and update operations. This is for improving the flow of the code and increasing readability, and because the aks operator only needs to pull and merge in data from the upstream AKS cluster when a provisioned or imported cluster is updated. When a cluster is created, it only needs to create a new `ManagedCluster` object if it does not already exist.

* Creating a cluster: When creating an AKS cluster, the aks operator pulls the cluster state and if it does not exist, creates a new `ManagedCluster` object with fields from the config spec.

* Updating a cluster: When updating an AKS cluster, the aks operator pulls the cluster state into a `ManagedCluster` object, then updates that object with fields from the config spec that were set/updated by Rancher or user-defined in Rancher. This approach enables us to only care about the fields that Rancher manages instead of having to maintain the logic every time that Azure adds a new field to their configuration. This also does not break the principal from the [KEv2 docs](https://github.com/rancher/rancher/wiki/Kontainer-Engine-v2:-The-new-way-rancher-integrates-with-cloud-providers#how-does-rancher-integrate-with-the-kev2-operators) of having non nil flags for managed fields in Rancher and nil flags for non-managed fields in Rancher. This piece just merges in the upstream state before an update to AKS so nil flags in Rancher don't overwrite fields in upstream (KEv2 thing: when I say upstream I mean the downstream cluster). 

* Authorized IP range bug: When building the upstream AKS cluster spec, the aks operator will check the upstream state and set authorized IP ranges correctly.

* Added missing func comments
 
## Testing
 

This issue requires a lot of testing because it touches all parts of provisioning an AKS cluster. Here's the basic test plan I used to smoke test my changes.

Test plan
1. Provisioned (via the Rancher UI kubernetes v1.23.5)
	a. 1 node AKS cluster
	b. 1 node AKS. Update auth IP range to 0.0.0.0/0 in Azure. Check that no config was overwritten in AKS/change was propogated as expected
	c. 1 node AKS cluster. Update HTTP Application Routing in rancher UI. Check that no config was overwritten in AKS
2. Import
	a. 3 node AKS cluster created via CLI with UDR and firewall. Check that no config was overwritten in AKS (agent pool, network). Update kubernetes version for the cluster configuration
	b. 3 node AKS cluster created via Azure portal

### DIFF
--- a/controller/aks-cluster-config-handler.go
+++ b/controller/aks-cluster-config-handler.go
@@ -244,7 +244,7 @@ func (h *Handler) createCluster(config *aksv1.AKSClusterConfig) (*aksv1.AKSClust
 		return config, err
 	}
 
-	err = aks.CreateOrUpdateCluster(ctx, credentials, resourceClusterClient, &config.Spec, config.Status.Phase)
+	err = aks.CreateCluster(ctx, credentials, resourceClusterClient, &config.Spec, config.Status.Phase)
 	if err != nil {
 		return config, fmt.Errorf("error failed to create cluster: %v ", err)
 	}
@@ -343,6 +343,7 @@ func (h *Handler) checkAndUpdate(config *aksv1.AKSClusterConfig) (*aksv1.AKSClus
 	if err != nil {
 		return config, err
 	}
+
 	return h.updateUpstreamClusterState(ctx, h.secretsCache, h.secrets, config, upstreamSpec)
 }
 
@@ -564,7 +565,8 @@ func GetClusterKubeConfig(ctx context.Context, secretsCache wranglerv1.SecretCac
 	return config, nil
 }
 
-// BuildUpstreamClusterState creates AKSClusterConfigSpec from existing cluster configuration
+// BuildUpstreamClusterState creates an AKSClusterConfigSpec (spec for the AKS cluster state) from the existing
+// cluster configuration.
 func BuildUpstreamClusterState(ctx context.Context, secretsCache wranglerv1.SecretCache, secretClient wranglerv1.SecretClient, spec *aksv1.AKSClusterConfigSpec) (*aksv1.AKSClusterConfigSpec, error) {
 	upstreamSpec := &aksv1.AKSClusterConfigSpec{}
 
@@ -687,7 +689,7 @@ func BuildUpstreamClusterState(ctx context.Context, secretsCache wranglerv1.Secr
 		if clusterState.APIServerAccessProfile.EnablePrivateCluster != nil {
 			upstreamSpec.PrivateCluster = clusterState.APIServerAccessProfile.EnablePrivateCluster
 		}
-		if clusterState.APIServerAccessProfile.AuthorizedIPRanges != nil {
+		if clusterState.APIServerAccessProfile.AuthorizedIPRanges != nil && *clusterState.APIServerAccessProfile.AuthorizedIPRanges != nil {
 			upstreamSpec.AuthorizedIPRanges = clusterState.APIServerAccessProfile.AuthorizedIPRanges
 		}
 	}
@@ -797,7 +799,7 @@ func (h *Handler) updateUpstreamClusterState(ctx context.Context, secretsCache w
 				clusterSpecCopy.NodePools = append(clusterSpecCopy.NodePools, n)
 			}
 		}
-		err = aks.CreateOrUpdateCluster(ctx, credentials, resourceClusterClient, clusterSpecCopy, config.Status.Phase)
+		err = aks.UpdateCluster(ctx, credentials, resourceClusterClient, clusterSpecCopy, config.Status.Phase)
 		if err != nil {
 			return config, fmt.Errorf("failed to update cluster: %v", err)
 		}
--- a/pkg/aks/create.go
+++ b/pkg/aks/create.go
@@ -9,6 +9,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2019-10-01/resources"
 	"github.com/Azure/go-autorest/autorest/to"
 	aksv1 "github.com/rancher/aks-operator/pkg/apis/aks.cattle.io/v1"
+	"github.com/sirupsen/logrus"
 )
 
 func CreateResourceGroup(ctx context.Context, groupsClient *resources.GroupsClient, spec *aksv1.AKSClusterConfigSpec) error {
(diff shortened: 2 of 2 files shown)


pkg/aks/create.go
```diff
@@ -148,17 +148,23 @@ func CreateOrUpdateCluster(ctx context.Context, cred *Credentials, clusterClient
...
-			AgentPoolProfiles: &agentPoolProfiles,
-			LinuxProfile:      linuxProfile,
-			NetworkProfile:    networkProfile,
-			AddonProfiles:     addonProfiles,
-		},
+	// Check if a managedCluster already exists in Azure. If so, then pull the configuration for that cluster instead
+	// of creating a brand new one.
+	managedCluster, err := clusterClient.Get(ctx, spec.ResourceGroup, spec.ClusterName)
+
+	if err != nil {
```
snasovich, line 155: Discussed this offline. 2 items to address:
- Improve error handling here, we cannot do the same here regardless of kind of error we get here.
- With setting of certain fields like `ManagedClusterProperties.KubernetesVersion` moved to under `if` here, these will no longer be updated in case we didn't get an error on `clusterClient.Get` and that would result in a regression where changes done to these fields in Rancher would not propagate down to Azure.

annablender: * Improved error handling
* Right-this data had to be merged together so as to allow changes in Rancher to propagate to Azure without overwriting other pre-existing values in Azure. I refactored most of `create.go` to separate out the Create and Update functionality (still in progress but what is mentioned above here is fixed. I can add specifically adding `KubernetesVersion` to my testing)

annablender: Apologies for the earlier force push. I've added a new commit for my changes today

pkg/aks/create.go
```diff
@@ -23,17 +23,134 @@ func CreateResourceGroup(ctx context.Context, groupsClient *resources.GroupsClie
...
+func UpdateCluster(ctx context.Context, cred *Credentials, clusterClient *containerservice.ManagedClustersClient,
+	spec *aksv1.AKSClusterConfigSpec, phase string) error {
+
+	// Create a new managed cluster from the AKS cluster config
+	managedCluster, err := CreateManagedCluster(ctx, cred, spec, phase)
+	if err != nil {
+		return err
+	}
+
+	// Pull the upstream cluster state. If the upstream cluster couldn't be found (not expected), then create a new
```
snasovich, line 57: Is this actually a valid situation where cluster is not found in AKS here?
I'd argue we should return an error even if it wasn't found as it's "not expected" rather than proceed forward with creating a new one.
I'm happy to hear arguments for doing it the way it's currently written.

pkg/aks/create.go
```diff
@@ -23,17 +23,134 @@ func CreateResourceGroup(ctx context.Context, groupsClient *resources.GroupsClie
...
+		if strings.Contains(err.Error(), spec.ClusterName + " was not found") {
+			return CreateCluster(ctx, cred, clusterClient, spec, phase)
+		} else {
+			return err
+		}
+	}
+
+	// Upstream cluster state was successfully pulled. Merge in updates without overwriting upstream fields: we never
+	// want to overwrite preconfigured values in Azure with nil values. So if a field on the managed cluster is nil,
+	// set it to the upstream value.
```
snasovich, line 70: It looks to me that the current approach is "build the AKS cluster object based on everything Rancher knows about (`managedCluster`) and set all the fields Rancher doesn't know about from object we got from AKS (`upstreamCluster`)". I think flipping this around makes more sense - get current object from AKS and update all the fields Rancher knows about on it before updating.
The big downside of the current approach is added maintenance - any time new (top-level) AKS field is added, we need to update this logic. I think it's much more natural to handle what Rancher supports ("knows about") instead of the reverse ("we don't know how much we don't know" type of situation).

pkg/aks/create.go
```diff
@@ -23,17 +23,134 @@ func CreateResourceGroup(ctx context.Context, groupsClient *resources.GroupsClie
...
+
+		// todo: use the AgentPoolProfiles iterator to pull agent pool changes from upstream before an update
+		//if managedCluster.AgentPoolProfiles == nil {
+		//	for profile := range managedCluster.AgentPoolProfiles {
+		//
+		//	}
+		//}
+
+		// linux profile
+		if managedCluster.LinuxProfile == nil {
```
snasovich, line 119: Don't we actually support `LinuxProfile` on Rancher side? So overwriting it from data we got from AKS (upstream) cluster seems we're giving precedence to it in the function that is supposed to propagate Rancher updates to AKS.

Again, goes in line with previous comment about handling fields we support in Rancher instead of handling the fields we don't support.

snasovich (changes requested):

pkg/aks/create.go
annablender: Ahh yes.. on second thought I'm with you here. I ran into some minor complications trying to figure out which upstream fields in Azure I should check for on the AKS cluster object (`managedCluster`). If we do it that we, we won't have to monitor a list of upstream fields that MAY be set.

annablender:
Hi all, these changes have been tested but I'd like some initial feedback on the design/location of code fix before I mark it ready for review. Thanks in advance!

pkg/aks/create.go
```diff
@@ -23,17 +24,162 @@ func CreateResourceGroup(ctx context.Context, groupsClient *resources.GroupsClie
...
+
+	// Upstream cluster state was successfully pulled. Merge in updates without overwriting upstream fields: we never
+	// want to overwrite preconfigured values in Azure with nil values. So only update fields pulled from AKS with
+	// values from the managed cluster if they are non nil.
+
+	/*
+	The following fields are managed in Rancher but are NOT configurable on update
+	- Name
+	- Location
+	- KubernetesVersion
```
cmurphy, line 73: This is configurable on update. Updating the k8s version of a cluster is a core feature for Rancher. I would double check on the others too.

pkg/aks/create.go
```diff
@@ -23,17 +24,162 @@ func CreateResourceGroup(ctx context.Context, groupsClient *resources.GroupsClie
...
+	spec *aksv1.AKSClusterConfigSpec, phase string) error {
+
+	// Create a new managed cluster from the AKS cluster config
+	managedCluster, err := CreateManagedCluster(ctx, cred, spec, phase)
+	if err != nil {
+		return err
+	}
+
+	// Pull the upstream cluster state
+	clusterUpdate, err := clusterClient.Get(ctx, spec.ResourceGroup, spec.ClusterName)
```
cmurphy, line 59: The actual state of the cluster in the provide is already available in the calling function in the controller, so there is no need to retrieve it again https://github.com/rancher/aks-operator/blob/a32c2bf0c3254301d1257345d4db786fef49162e/controller/aks-cluster-config-handler.go#L702

A lot of the logic for determining what fields to update and what fields are unmanaged is already contained in that updateUpstreamClusterState method. Can you clarify what is different here, and why it is needed? Is this following the principle of using nil as a flag for what fields are managed and unmanaged? https://github.com/rancher/rancher/wiki/Kontainer-Engine-v2:-The-new-way-rancher-integrates-with-cloud-providers

annablender: Good catch, but what if there's been an update in AKS since the upstreamSpec was last built?

Sure - that's true. For a lot of `updateUpstreamClusterState,` we just compare with the upstream state and then update upstream with whatever data is in the config spec. Thus, if a field in Rancher is nil and the upstream value is not the upstream value often gets overwritten. `updateCluster` in `create.go` addresses that by merging in data from upstream before the update is done. I could probably add that logic to `updateUpstreamClusterState` but that may make the operator less readable. Although it would leave just the create logic left in `create.go` which I actually might like better.. pending discussion

controller/aks-cluster-config-handler.go
```diff
@@ -687,7 +688,7 @@ func BuildUpstreamClusterState(ctx context.Context, secretsCache wranglerv1.Secr
 		if clusterState.APIServerAccessProfile.EnablePrivateCluster != nil {
 			upstreamSpec.PrivateCluster = clusterState.APIServerAccessProfile.EnablePrivateCluster
 		}
-		if clusterState.APIServerAccessProfile.AuthorizedIPRanges != nil {
+		if &clusterState.APIServerAccessProfile.AuthorizedIPRanges != nil {
```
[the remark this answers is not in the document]
snasovich: Do we actually need to fix anything here? Given Colleen's comment I don't see anything wrong with the current code.
@annablender , what is the exact issue you ran into here?

pkg/aks/create.go
snasovich, line 59: `clusterUpdate` is probably not the best name here. This is not really an "update", so maybe `updatedCluster` or even better `aksCluster` so it's clear it's AKS object?

pkg/aks/create.go
```diff
@@ -23,17 +24,162 @@ func CreateResourceGroup(ctx context.Context, groupsClient *resources.GroupsClie
...
+	if err != nil {
+		logrus.Errorf("Error getting upstream cluster [%s]: %s", spec.ClusterName, err.Error())
```
snasovich, line 61:
```suggestion
		logrus.Errorf("Error getting upstream AKS cluster by name [%s]: %s", spec.ClusterName, err.Error())
```

pkg/aks/create.go
snasovich: Seems this comment and code below is a bit contradicting - we're saying these fields cannot be updated, but then we're clearly updating them based on `managedCluster` (coming from Rancher).

controller/aks-cluster-config-handler.go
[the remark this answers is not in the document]
annablender: Thank you @cmurphy I meant to dereference. @snasovich I was not seeing the authorized IP range value/s displayed correctly in the UI

rmweir (dismissed):
LGTM

pkg/aks/create.go
```diff
@@ -23,17 +24,159 @@ func CreateResourceGroup(ctx context.Context, groupsClient *resources.GroupsClie
...
+	The following fields are managed in Rancher but are NOT configurable on update
+	- Name
+	- Location
+	- DNSPrefix
+	- EnablePrivateCluster
+	- LoadBalancerProfile
+	 */
+
+	// Update kubernetes version
+	aksCluster.KubernetesVersion = managedCluster.KubernetesVersion
```
cmurphy, line 79: If a cluster is imported, we might not have kubernetes version set on the spec in Rancher. Is there a risk this could override the upstream value with a nil value?

pkg/aks/create.go
```diff
@@ -23,17 +24,159 @@ func CreateResourceGroup(ctx context.Context, groupsClient *resources.GroupsClie
...
+
+	// Add/update agent pool profiles
+	for _, ap := range *managedCluster.AgentPoolProfiles {
+		if !hasAgentPoolProfile(ap.Name, managedCluster.AgentPoolProfiles) {
```
cmurphy, line 83: I'm a little confused by this logic. Is one of these supposed to be `aksCluster` instead of `managedCluster`?

pkg/aks/create.go
```diff
@@ -177,22 +325,19 @@ func CreateOrUpdateCluster(ctx context.Context, cred *Credentials, clusterClient
...
-		managedCluster,
-	)
-
-	return err
+	return managedCluster, nil
 }
 
+
+// CreateOrUpdateAgentPool creates a new pool(s) in AKS. If one already exists it updates the upstream node pool with
+// any provided updates.
```
cmurphy, line 340: love the added godocs :heart:

cmurphy (changes requested):
I think there are some minor errors in UpdateCluster, otherwise lgtm

jakefhyde (changes requested):
LGTM, Requested Changes are more for clarification on a few things.

pkg/aks/create.go
```diff
@@ -23,17 +24,159 @@ func CreateResourceGroup(ctx context.Context, groupsClient *resources.GroupsClie
...
+	if managedCluster.APIServerAccessProfile != nil && managedCluster.APIServerAccessProfile.AuthorizedIPRanges != nil {
+
+		for index := range *managedCluster.APIServerAccessProfile.AuthorizedIPRanges {
+			ipRange := (*managedCluster.APIServerAccessProfile.AuthorizedIPRanges)[index]
+
+			if !hasAuthorizedIpRange(ipRange, aksCluster.APIServerAccessProfile.AuthorizedIPRanges) {
+				*aksCluster.APIServerAccessProfile.AuthorizedIPRanges = append(*aksCluster.APIServerAccessProfile.AuthorizedIPRanges, ipRange)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
jakefhyde: Okay, max 144 checks isn't too scary then.

jakefhyde (approved):

rmweir (approved):

cmurphy (commented):
lgtm, can you squash the commits?

cmurphy (approved):

jakefhyde (approved):
LGTM

PR #60, opened 2022-06-22
update dep to k8s v1.24

kinarashah:
https://github.com/rancher/rancher/issues/37711


go.mod
```diff
@@ -1,23 +1,94 @@
 module github.com/rancher/aks-operator
 
-go 1.16
+go 1.17
 
-replace k8s.io/client-go => k8s.io/client-go v0.21.2
+replace (
+	github.com/rancher/lasso => github.com/kinarashah/lasso v0.0.0-20220524091619-0f0b92e5262c
```
jiaqiluo, line 6: Should we switch back to rancher/lasso?  I guess that means we need to update rancher/lasso first, right?

jiaqiluo (changes requested):

jiaqiluo (approved):

PR #65, opened 2022-08-31
Add duplicate cluster check

annablender:
Issue: https://github.com/rancher/rancher/issues/37955

## Problem

When you try to provision an AKS cluster with the same name and resource group as one that already exists in Azure, an error appears in the rancher UI because the [CreateOrUpdate call to AKS](https://github.com/rancher/aks-operator/blob/a32c2bf0c3254301d1257345d4db786fef49162e/pkg/aks/create.go#L186) fails.

## Solution

Added check for existing cluster in the aks operator `CreateCluster` function. If the cluster name in the specified resource group already exists in Azure, an error `cluster [name] already exists in Azure. Please import it` will now display in the rancher UI. This is much clearer for the customer and short circuits unnecessary API calls being made to create the managed cluster.

## Testing

Smoke test
* Create an AKS cluster in Azure Portal with [name] and [resource group]
* Create an AKS cluster in the rancher UI with [name] and [resource group]. Verify the following error thrown:

image

* Create an AKS cluster in the rancher UI with [name] and [resource group 2]. Verify cluster provisions successfully
* Create an AKS cluster in the rancher UI with [name2] and [resource group]. Verify cluster provisions successfully

### DIFF
--- a/controller/aks-cluster-config-handler.go
+++ b/controller/aks-cluster-config-handler.go
@@ -221,6 +221,17 @@ func (h *Handler) createCluster(config *aksv1.AKSClusterConfig) (*aksv1.AKSClust
 		return config, err
 	}
 
+	resourceClusterClient, err := aks.NewClusterClient(credentials)
+	if err != nil {
+		return config, err
+	}
+
+	logrus.Infof("Checking if cluster [%s] exists", config.Spec.ClusterName)
+
+	if aks.ExistsCluster(ctx, resourceClusterClient, &config.Spec) {
+		return config, fmt.Errorf("cluster [%s] already exists in AKS. Please import it", config.Spec.ClusterName)
+	}
+
 	resourceGroupsClient, err := aks.NewResourceGroupClient(credentials)
 	if err != nil {
 		return config, err
@@ -238,11 +249,6 @@ func (h *Handler) createCluster(config *aksv1.AKSClusterConfig) (*aksv1.AKSClust
 	}
 
 	logrus.Infof("Creating AKS cluster [%s]", config.Spec.ClusterName)
-
-	resourceClusterClient, err := aks.NewClusterClient(credentials)
-	if err != nil {
-		return config, err
-	}
 
 	err = aks.CreateCluster(ctx, credentials, resourceClusterClient, &config.Spec, config.Status.Phase)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


controller/aks-cluster-config-handler.go
```diff
@@ -221,6 +221,17 @@ func (h *Handler) createCluster(config *aksv1.AKSClusterConfig) (*aksv1.AKSClust
...
 	}
 
+	resourceClusterClient, err := aks.NewClusterClient(credentials)
+	if err != nil {
+		return config, err
+	}
+
+	logrus.Infof("Checking if cluster [%s] exists", config.Spec.ClusterName)
+
+	if aks.ExistsCluster(ctx, resourceClusterClient, &config.Spec) {
```
jakefhyde, line 231: I think we may want to actually update the logic in `aks.ExistsCluster`. Just by looking at it:
```
// ExistsCluster Check if AKS managed Kubernetes cluster exist
func ExistsCluster(ctx context.Context, clusterClient *containerservice.ManagedClustersClient, spec *aksv1.AKSClusterConfigSpec) bool {
	resp, err := clusterClient.Get(ctx, spec.ResourceGroup, spec.ClusterName)

	return err == nil && resp.StatusCode == 200
}
What if the user doesn't have permissions to list clusters, but the cluster exists? IMO we should either return `(bool, error)` from that function and just require that users have list permissions in our docs for AKS, or if we can't do that we should be very explicit about logging the error and possibly bubble it up. I doubt we could get a scenario where `err == nil` and the response code is up in the 500s though. I think it's very likely that it either returns 404 for both lack of permissions and object not existing, or just 403 in the former case. 
```

controller/aks-cluster-config-handler.go
```diff
@@ -221,6 +221,17 @@ func (h *Handler) createCluster(config *aksv1.AKSClusterConfig) (*aksv1.AKSClust
...
+		return config, fmt.Errorf("cluster [%s] already exists in AKS. Please import it", config.Spec.ClusterName)
```
jakefhyde, line 232:
```suggestion
		return config, fmt.Errorf("cluster [%s] already exists in AKS", config.Spec.ClusterName)
```

I think we might not want to suggest importing the cluster, if a user is trying to provision a hosted cluster from rancher they likely want rancher to manage the lifecycle.

jakefhyde (changes requested):

controller/aks-cluster-config-handler.go
annablender: Are you saying that we shouldn't assume the user's end goal in the error msg?

jakefhyde (approved):

jiaqiluo (approved):
LGTM

PR #72, opened 2022-12-09
Add new unit tests.

alexander-demicev:
In order to increase the maintainability of the operator we need to have more unit tests, this PR includes a basic setup for writing tests:
- All azure services are now in a separate package.
- Every service has a generated mock, we can now test what is actually passed to the azure API.
- I added a new github workflow for running unit tests.


pkg/aks/check.go
```diff
@@ -8,6 +8,7 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/services/operationalinsights/mgmt/2020-08-01/operationalinsights"
 	"github.com/Azure/go-autorest/autorest/to"
+	"github.com/rancher/aks-operator/pkg/aks/services"
```
annablender, line 11: From what I can tell, @alexander-demicev you are extracting the `*containerservices` calls and switching to use the Azure client functions defined in the new `services` package. Seems fine design wise, I just want to make sure there are no functional gaps there so both provisioned/provisioned cluster update and an imported Azure cluster should be tested

pkg/aks/create_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/golang/mock/gomock"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/rancher/aks-operator/pkg/aks/services/mock_services"
+	aksv1 "github.com/rancher/aks-operator/pkg/apis/aks.cattle.io/v1"
+)
+
+var _ = Describe("CreateResourceGroup", func() {
+	var (
+		mockCtrl          *gomock.Controller
```
annablender, line 20: IMO I prefer variables to be written out so it's completely readable, like `mockController` and `mockResourceGroupClient`

pkg/aks/suite_test.go
```diff
@@ -0,0 +1,18 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"golang.org/x/net/context"
+)
+
+var (
+	ctx = context.Background()
+)
+
+func TestAPIs(t *testing.T) {
```
annablender, line 15: Why is this test file here if it's this small?

.github/workflows/unit.yaml
```diff
@@ -12,7 +12,7 @@ jobs:
     - name: Install Go
       uses: actions/setup-go@v3
       with:
-        go-version: 1.17.x
+        go-version: 1.18.x
```
annablender, line 15: I believe we are on go 1.19 for rancher now. I use 1.19.3. @kinarashah should this be bumped for the operators?

pkg/aks/check_test.go
```diff
@@ -24,3 +31,83 @@ func Test_generateUniqueLogWorkspace(t *testing.T) {
 		assert.Len(t, got, 63)
 	}
 }
+
+var _ = Describe("checkLogAnalyticsWorkspaceForMonitoring", func() {
+	var (
+		workplacesClientMock *mock_services.MockWorkplacesClientInterface
+		mockCtrl             *gomock.Controller
```
annablender, line 38: Same as above, `mockController`

pkg/aks/create_test.go
```diff
@@ -56,3 +58,381 @@ var _ = Describe("CreateResourceGroup", func() {
...
+		mockCtrl             *gomock.Controller
+		workplacesClientMock *mock_services.MockWorkplacesClientInterface
+		clusterSpec          *aksv1.AKSClusterConfigSpec
+		cred                 *Credentials
+	)
+
+	BeforeEach(func() {
+		mockCtrl = gomock.NewController(GinkgoT())
+		workplacesClientMock = mock_services.NewMockWorkplacesClientInterface(mockCtrl)
+		clusterSpec = &aksv1.AKSClusterConfigSpec{
```
annablender, line 73: Can you see if you can define this as json in an embedded json test file and then `json.Unmarshal` it into an `aksv1.AKSClusterConfigSpec` object?

pkg/aks/create_test.go
```diff
@@ -56,3 +58,381 @@ var _ = Describe("CreateResourceGroup", func() {
...
+		Expect(managedCluster.APIServerAccessProfile).ToNot(BeNil())
+		Expect(managedCluster.APIServerAccessProfile.AuthorizedIPRanges).ToNot(BeNil())
+		ipRanges := *managedCluster.APIServerAccessProfile.AuthorizedIPRanges
+		clusterSpecIPRanges := *clusterSpec.AuthorizedIPRanges
+		Expect(ipRanges).To(HaveLen(1))
+		Expect(ipRanges[0]).To(Equal(clusterSpecIPRanges[0]))
+		Expect(managedCluster.APIServerAccessProfile.EnablePrivateCluster).To(Equal(clusterSpec.PrivateCluster))
+	})
+
+	It("should succefully create managed cluster with custom load balancer sku", func() {
```
annablender, line 184: Fix to be `successfully` here and all other instances

pkg/aks/create.go
```diff
@@ -39,219 +38,105 @@ func CreateCluster(ctx context.Context, cred *Credentials, clusterClient service
...
-	return err
-}
-
-// createManagedCluster creates a new managed Kubernetes cluster.
-func createManagedCluster(ctx context.Context, cred *Credentials, workplacesClient services.WorkplacesClientInterface, spec *aksv1.AKSClusterConfigSpec, phase string) (containerservice.ManagedCluster, error) {
-	managedCluster := containerservice.ManagedCluster{}
-
-	// Get tags from config spec
-	tags := make(map[string]*string)
+	managedCluster.Tags = make(map[string]*string)
```
annablender, line 57: This will cause a panic if `managedCluster.Tags` is nil and a key tries to get set.

In addition, if you remove

`managedCluster.Tags = make(map[string]*string)`

the controller will see that `managedCluster.Tags is nil` and will try to overwrite the Azure field with that state. I think you want to leave this as an empty map object even if no tags are added so that upstream tags are not removed on creation.

annablender (changes requested):

annablender:
@alexander-demicev In general, what you've done is great.

* I would've liked the refactoring to be in a separate PR so it's not as much of a beast to review, but if that can't be done no worries. Please dev test create/update an AKS cluster via the rancher ui and make sure no fields are missing or getting overwritten in upstream. That has been a concern with some controller refactoring that I've done before so it's worth making sure there are no regressions
* Please add instructions on how to run the unit tests here https://github.com/rancher/aks-operator/blob/master/README.md

pkg/aks/suite_test.go
alexander-demicev: This PR introduces `ginkgo` as testing framework and it requires initialization https://onsi.github.io/ginkgo/#bootstrapping-a-suite, this code will bootstrap the test runner and start tests.

.github/workflows/unit.yaml
alexander-demicev: +1 for updating to 1.19 if possible

pkg/aks/create_test.go
alexander-demicev: I'd like to keep it as is since we don't reuse this cluster spec in other tests. It should be easier to see what values are set this way.

alexander-demicev:
@annablender I did some manual testing of the PR. Actually I wrote unit tests before refactoring the logic, the behavior should be the same. I also opened a PR with a setup for e2e tests https://github.com/rancher/aks-operator/pull/74, in future we will have more confidence that no regressions are introduced. 
Is it ok if I update readme in a follow-up PR?

pkg/aks/create.go
```diff
@@ -38,219 +38,106 @@ func CreateCluster(ctx context.Context, cred *Credentials, clusterClient *contai
...
-		NetworkPlugin:   containerservice.Kubenet,
-		NetworkPolicy:   containerservice.NetworkPolicy(to.String(spec.NetworkPolicy)),
-		LoadBalancerSku: containerservice.Standard,
+		NetworkPolicy: containerservice.NetworkPolicy(to.String(spec.NetworkPolicy)),
 	}
 
-	if spec.LoadBalancerSKU != nil {
+	networkProfile.NetworkPlugin = containerservice.Kubenet
+	if to.String(spec.NetworkPlugin) != "" {
+		networkProfile.NetworkPlugin = containerservice.NetworkPlugin(to.String(spec.NetworkPlugin))
```
mjura, line 71: I think here we should rely on const values, there are only two possible `containerservice.Kubenet` and `containerservice.Azure`. Converting it from config string can cause some problems in future

pkg/aks/create.go
```diff
@@ -38,219 +38,106 @@ func CreateCluster(ctx context.Context, cred *Credentials, clusterClient *contai
...
+	}
+
+	networkProfile.LoadBalancerSku = containerservice.Standard
+	if to.String(spec.LoadBalancerSKU) != "" {
 		networkProfile.LoadBalancerSku = containerservice.LoadBalancerSku(to.String(spec.LoadBalancerSKU))
```
mjura, line 76: Again please use const `containerservice.Standard`

pkg/aks/create.go
```diff
@@ -38,219 +38,106 @@ func CreateCluster(ctx context.Context, cred *Credentials, clusterClient *contai
...
 	}
+	managedCluster.Tags = tags
 
-	// Get network profile from config spec
-	var vmNetSubnetID *string
 	networkProfile := &containerservice.NetworkProfile{
-		NetworkPlugin:   containerservice.Kubenet,
-		NetworkPolicy:   containerservice.NetworkPolicy(to.String(spec.NetworkPolicy)),
-		LoadBalancerSku: containerservice.Standard,
+		NetworkPolicy: containerservice.NetworkPolicy(to.String(spec.NetworkPolicy)),
```
mjura, line 66: About `NetworkPolicy` in Azure there is limitation that `containerservice.NetworkPolicyCalico` can be used only with `containerservice.Kubenet`. When `containerservice.Kubenet` is used then NetworkPolicy can be `None` or `containerservice.NetworkPolicyCalico`

mjura: For `NetworkPlugin == containerservice.Azure` NetworkPolicy can be `None` or `containerservice.NetworkPolicyCalico` or `containerservice.NetworkPolicyAzure`

pkg/aks/create.go
```diff
@@ -38,219 +38,106 @@ func CreateCluster(ctx context.Context, cred *Credentials, clusterClient *contai
...
 		}
 	}
+	managedCluster.Tags = tags
 
-	// Get network profile from config spec
-	var vmNetSubnetID *string
 	networkProfile := &containerservice.NetworkProfile{
-		NetworkPlugin:   containerservice.Kubenet,
-		NetworkPolicy:   containerservice.NetworkPolicy(to.String(spec.NetworkPolicy)),
-		LoadBalancerSku: containerservice.Standard,
```
mjura, line 187: this should be always `LoadBalancerSku: containerservice.Standard`

pkg/aks/create.go
mjura: And this should be set always as `networkProfile.LoadBalancerSku = containerservice.Standard`

pkg/aks/create.go
```diff
@@ -38,219 +38,106 @@ func CreateCluster(ctx context.Context, cred *Credentials, clusterClient *contai
...
 		if np.AvailabilityZones != nil && len(*np.AvailabilityZones) > 0 {
 			agentProfile.AvailabilityZones = np.AvailabilityZones
 		}
 
-		if np.EnableAutoScaling != nil && *np.EnableAutoScaling {
+		if to.StringSlice(np.AvailabilityZones) != nil {
+			agentProfile.AvailabilityZones = np.AvailabilityZones
+		}
+
+		if to.Bool(np.EnableAutoScaling) {
```
mjura, line 119: It has to be like this `if np.EnableAutoScaling != nil && *np.EnableAutoScaling` because of managed fields by Rancher

pkg/aks/create.go
```diff
@@ -263,84 +150,67 @@ func CreateManagedCluster(ctx context.Context, cred *Credentials, spec *aksv1.AK
...
-
 	if phase != "updating" && phase != "active" {
 		managedCluster.ServicePrincipalProfile = &containerservice.ManagedClusterServicePrincipalProfile{
 			ClientID: to.StringPtr(cred.ClientID),
 			Secret:   to.StringPtr(cred.ClientSecret),
 		}
 	}
 
-	if spec.DNSPrefix != nil {
+	if to.String(spec.DNSPrefix) != "" {
```
mjura, line 190: Check first `if spec.DNSPrefix != nil {`

mjura (commented):
Great job @alexander-demicev and thank you for doing this, I am for mostly all changes, but changes to logic for `newManagedCluster` and `updateCluster` should be done in another PR with a lot of testing. This might brake as I know Azure Cloud.

pkg/aks/create.go
alexander-demicev: I was trying to follow the current logic and we have this code there https://github.com/rancher/aks-operator/blob/master/pkg/aks/create.go#L191. If we can only use standard LBs should this condition be removed?

pkg/aks/create.go
alexander-demicev: +1 for this change, I just followed the current logic https://github.com/rancher/aks-operator/blob/master/pkg/aks/create.go#L195

pkg/aks/create.go
alexander-demicev: shouldn't `to.Bool()` be the same https://github.com/Azure/go-autorest/blob/main/autorest/to/convert.go#L72 or I'm missing something?

pkg/aks/create.go
alexander-demicev: `to.String()` will always return empty string if the value is nil https://github.com/Azure/go-autorest/blob/main/autorest/to/convert.go#L26

alexander-demicev (commented):
@mjura I wrote tests before making changes to these functions but I have no objections about splitting this PR into smaller ones if that's a concern. Thanks for reviewing it.

alexander-demicev:
Closing it in favor of a smaller PRs https://github.com/rancher/aks-operator/pull/75

PR #76, opened 2023-01-06
Pass node pool's subnet id to azure client

alexander-demicev:
Currently, node pool's subnet id is not set when the operator updates a node pool. This makes azure client try to delete it if the value is set to nil. The PR should fix by passing this value if possible.

Fixes: https://github.com/rancher/rancher/issues/40087

### DIFF
--- a/charts/aks-operator-crd/templates/crds.yaml
+++ b/charts/aks-operator-crd/templates/crds.yaml
@@ -122,6 +122,9 @@ spec:
                     vmSize:
                       nullable: true
                       type: string
+                    vnetSubnetID:
+                      nullable: true
+                      type: string
                   type: object
                 nullable: true
                 type: array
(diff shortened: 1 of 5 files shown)


mjura (approved):
I have tested it and it seems it resolve issue https://github.com/rancher/rancher/issues/40087

mjura:
@alexander-demicev Thank you so much for preparing this fix. Great job!!!

@annablender We are modifying `aks-operator-crd` helm chart and fix will be shipped with `aks-operator`. Could you please help us with releasing this bug fix.

mjura:
I have reported also Rancher issue to track this bug fix https://github.com/rancher/rancher/issues/40087

Please add it to commit message.

jakefhyde (approved):

richardcase (approved):

annablender (approved):
LGTM this can be merged

PR #77, opened 2023-01-06
Check nil value for aksCluster AddonProfiles

mjura:
For changing AddonProfiles for imported AKS cluster, we have to check first if AddonProfiles for aksCluster spec is not nil. If yes then it has to be initialized.

Fixes: https://github.com/rancher/rancher/issues/40087
Signed-off-by: Michal Jura <mjura@suse.com>


alexander-demicev (approved):

richardcase (approved):

annablender (approved):
This will only be for the imported case. Technically, you don't have to create the map here for `aksCluster.AddonProfiles` to get set correctly with profiles from upstream but I don't think it will cause a regression. LGTM thank you

PR #78, opened 2023-01-10
Add more unit tests for azure services

alexander-demicev:
This PR adds more unit tests for azure services. There are no logical changes in cluster create/update actions.

### DIFF
--- a/pkg/aks/check.go
+++ b/pkg/aks/check.go
@@ -113,7 +113,7 @@ var regionToOmsRegionMap = map[string]string{
 	"chinanorth2": "chinaeast2",
 }
 
-func CheckLogAnalyticsWorkspaceForMonitoring(ctx context.Context, client services.WorkplacesClientInterface,
+func checkLogAnalyticsWorkspaceForMonitoring(ctx context.Context, client services.WorkplacesClientInterface,
 	location string, group string, wsg string, wsn string) (workspaceID string, err error) {
 
 	workspaceRegion, ok := regionToOmsRegionMap[location]
--- a/pkg/aks/check_test.go
+++ b/pkg/aks/check_test.go
@@ -1,8 +1,15 @@
 package aks
 
 import (
+	"errors"
 	"testing"
 
+	"github.com/Azure/azure-sdk-for-go/services/operationalinsights/mgmt/2020-08-01/operationalinsights"
+	"github.com/Azure/go-autorest/autorest/to"
+	"github.com/golang/mock/gomock"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/rancher/aks-operator/pkg/aks/services/mock_services"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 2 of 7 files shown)


pkg/aks/check.go
```diff
@@ -113,7 +113,7 @@ var regionToOmsRegionMap = map[string]string{
 	"chinanorth2": "chinaeast2",
 }
 
-func CheckLogAnalyticsWorkspaceForMonitoring(ctx context.Context, client services.WorkplacesClientInterface,
+func checkLogAnalyticsWorkspaceForMonitoring(ctx context.Context, client services.WorkplacesClientInterface,
```
mjura, line 116: `CheckLogAnalyticsWorkspaceForMonitoring` is public function, I was thinking that we want keep them starting with UpperCase

mjura (approved):
Looks ok, only small nitpick

pkg/aks/check.go
alexander-demicev: I noticed it's only used inside `aks` package so I thought we can make it private

alexander-demicev: I've just done a quick search across rancher org https://github.com/search?q=org%3Arancher+CheckLogAnalyticsWorkspaceForMonitoring&type=code
it should be safe to make it private

richardcase: Could we consider adding apidiff in the future to check for public api changes?

pkg/aks/check_test.go
```diff
@@ -24,3 +31,83 @@ func Test_generateUniqueLogWorkspace(t *testing.T) {
...
+		workplacesClientMock.EXPECT().CreateOrUpdate(ctx, gomock.Any(), gomock.Any(), gomock.Any()).Return(operationalinsights.WorkspacesCreateOrUpdateFuture{}, errors.New("error"))
+
+		_, err := checkLogAnalyticsWorkspaceForMonitoring(ctx,
+			workplacesClientMock,
+			"eastus", "workspaceResourceGroup", "", "workspaceName")
+
+		Expect(err).To(HaveOccurred())
+	})
+
+	It("should fail if AsyncCreateUpdateResult returns error", func() {
```
richardcase, line 102: Maybe in the future we could add a test where it fails initially but on the next poll it passes, a nice to have for the future.

pkg/aks/create_test.go
```diff
@@ -14,45 +17,422 @@ import (
...
+		mockController       *gomock.Controller
+		workplacesClientMock *mock_services.MockWorkplacesClientInterface
+		clusterSpec          *aksv1.AKSClusterConfigSpec
+		cred                 *Credentials
+	)
+
+	BeforeEach(func() {
+		mockController = gomock.NewController(GinkgoT())
+		workplacesClientMock = mock_services.NewMockWorkplacesClientInterface(mockController)
+		clusterSpec = &aksv1.AKSClusterConfigSpec{
```
richardcase, line 73: nit: might be nice to separate this out into a function.

richardcase (approved):
Looks good. Just 1 nit and a couple of comments.

pkg/aks/check.go
alexander-demicev: Yes, I can add apidiff check

pkg/aks/update.go
```diff
@@ -5,8 +5,161 @@ import (
...
+		updateCluster(aksCluster, managedCluster),
+	)
+
+	return err
+}
+
+func updateCluster(aksCluster containerservice.ManagedCluster, managedCluster containerservice.ManagedCluster) containerservice.ManagedCluster {
+	/*
+		The following fields are managed in Rancher but are NOT configurable on update
+		- Name
```
richardcase, line 51: If these are changed should we log a warning?

richardcase (approved):
LGTM.

One comment around fields that can't be changed on update.

PR #82, opened 2023-01-19
Add Nodes Labels, Taints and MaxSurge to AgentPool

mjura:
Fixes https://github.com/rancher/rancher/issues/40214

### DIFF
--- a/pkg/aks/create.go
+++ b/pkg/aks/create.go
@@ -109,8 +109,14 @@ func createManagedCluster(ctx context.Context, cred *Credentials, workplacesClie
 			VMSize:              containerservice.VMSizeTypes(np.VMSize),
 			Mode:                containerservice.AgentPoolMode(np.Mode),
 			OrchestratorVersion: np.OrchestratorVersion,
+			NodeLabels:          np.NodeLabels,
+			NodeTaints:          np.NodeTaints,
 		}
-
+		if np.MaxSurge != nil {
+			agentProfile.UpgradeSettings = &containerservice.AgentPoolUpgradeSettings{
+				MaxSurge: np.MaxSurge,
+			}
+		}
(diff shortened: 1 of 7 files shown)


mjura:
There is missing still part for creating AKS clusters from Rancher UI, only AKS imported clusters were covered

pkg/aks/create.go
```diff
@@ -233,6 +239,14 @@ func CreateOrUpdateAgentPool(ctx context.Context, agentPoolClient services.Agent
 		MinCount:            np.MinCount,
 		MaxCount:            np.MaxCount,
 		VnetSubnetID:        np.VnetSubnetID,
+		NodeLabels:          np.NodeLabels,
+		NodeTaints:          np.NodeTaints,
+	}
+
+	if np.MaxSurge != nil {
```
alexander-demicev, line 246: can you also a unit test for this?

pkg/aks/create.go
```diff
@@ -109,8 +109,14 @@ func createManagedCluster(ctx context.Context, cred *Credentials, workplacesClie
 			VMSize:              containerservice.VMSizeTypes(np.VMSize),
 			Mode:                containerservice.AgentPoolMode(np.Mode),
 			OrchestratorVersion: np.OrchestratorVersion,
+			NodeLabels:          np.NodeLabels,
+			NodeTaints:          np.NodeTaints,
+		}
+		if np.MaxSurge != nil {
```
alexander-demicev, line 115: can you add a unit test for these fields? Let's make sure we don't lose the functionality in future.

pkg/aks/create.go
mjura: I have added unit test

pkg/aks/create.go
mjura: I have added unit test

alexander-demicev (approved):

PR #84, opened 2023-01-22
Make Azure services compliant with Rancher

mjura:
For future implementation:
- change `BuildUpstreamClusterState` function definition and pass cluster client object instead of passing secrets and creating cluster client inside of it
- adjust Rancher integration  in https://github.com/rancher/rancher/blob/release/v2.7/pkg/controllers/management/clusterupstreamrefresher/aks_upstream_spec.go#L14 and use new function definition

### DIFF
--- a/controller/aks-cluster-config-handler.go
+++ b/controller/aks-cluster-config-handler.go
@@ -124,7 +124,7 @@ func (h *Handler) OnAksConfigChanged(_ string, config *aksv1.AKSClusterConfig) (
 		return nil, nil
 	}
 
-	if err := h.getClients(config); err != nil {
+	if err := h.getAzureClients(config); err != nil {
 		return config, fmt.Errorf("error getting Azure clients: %w", err)
 	}
 
@@ -143,7 +143,7 @@ func (h *Handler) OnAksConfigChanged(_ string, config *aksv1.AKSClusterConfig) (
 }
 
 func (h *Handler) OnAksConfigRemoved(_ string, config *aksv1.AKSClusterConfig) (*aksv1.AKSClusterConfig, error) {
-	if err := h.getClients(config); err != nil {
+	if err := h.getAzureClients(config); err != nil {
 		return config, fmt.Errorf("error getting Azure clients: %w", err)
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/aks/client.go
```diff
@@ -135,3 +137,27 @@ func GetCachedTenantID(secretClient secretClient, subscriptionID string, secret
 	}
 	return tenantID, err
 }
+
+func NewResourceGroupClient(cred *Credentials) (*resources.GroupsClient, error) {
```
alexander-demicev, line 141: is this method used anywhere?

controller/aks-cluster-config-handler.go
```diff
@@ -554,10 +562,18 @@ func (h *Handler) GetClusterKubeConfig(ctx context.Context, spec *aksv1.AKSClust
 
 // buildUpstreamClusterState creates an AKSClusterConfigSpec (spec for the AKS cluster state) from the existing
 // cluster configuration.
-func (h *Handler) BuildUpstreamClusterState(ctx context.Context, spec *aksv1.AKSClusterConfigSpec) (*aksv1.AKSClusterConfigSpec, error) {
+func BuildUpstreamClusterState(ctx context.Context, secretsCache wranglerv1.SecretCache, secretClient wranglerv1.SecretClient, spec *aksv1.AKSClusterConfigSpec) (*aksv1.AKSClusterConfigSpec, error) {
```
alexander-demicev, line 565: can we pass cluster client instead of creating it on every reconcile?

alexander-demicev (changes requested):
looks good, I have only a couple of questions

pkg/aks/client.go
mjura: yes, it can be removed

controller/aks-cluster-config-handler.go
mjura: yes, we can pass cluster client but it will break once again Rancher integration https://github.com/rancher/rancher/blob/release/v2.7/pkg/controllers/management/clusterupstreamrefresher/aks_upstream_spec.go#L14

I would keep it as it is for now, we want to unblock fixes for customer issue first and then we can continue with refactoring

alexander-demicev: We can create the client somewhere in core rancher and pass it to this function. Let's merge the PR now and refactor it when we release the fix.

alexander-demicev (approved):

annablender (approved):
LGTM, before the aforementioned refactoring

PR #85, opened 2023-01-23
Refactor create/update code to make it more readable

alexander-demicev:
Refactor create/update code to make it more readable. I also added new unit tests.

### DIFF
--- a/pkg/aks/create.go
+++ b/pkg/aks/create.go
@@ -10,6 +10,7 @@ import (
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/rancher/aks-operator/pkg/aks/services"
 	aksv1 "github.com/rancher/aks-operator/pkg/apis/aks.cattle.io/v1"
+	"github.com/sirupsen/logrus"
 )
 
 func CreateResourceGroup(ctx context.Context, groupsClient services.ResourceGroupsClientInterface, spec *aksv1.AKSClusterConfigSpec) error {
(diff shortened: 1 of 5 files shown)


pkg/aks/create.go
```diff
@@ -38,98 +39,123 @@ func CreateCluster(ctx context.Context, cred *Credentials, clusterClient service
...
+		networkProfile.NetworkPlugin = containerservice.Kubenet
+	default:
+		return nil, fmt.Errorf("networkPlugin '%s' is not supported", to.String(spec.NetworkPlugin))
+	}
+
+	switch to.String(spec.LoadBalancerSKU) { // TODO: only standard is supported for now, find a way to validate this
+	case string(containerservice.Standard):
+		networkProfile.LoadBalancerSku = containerservice.Standard
+	case string(containerservice.Basic):
+		logrus.Warnf("loadBalancerSKU 'basic' is not supported")
```
mjura, line 91: I will test it if it is possible to create AKS with `loadBalancerSKU = containerservice.Basic`

mjura (commented):
I have to check LoadBalancer = Basic

pkg/aks/create.go
alexander-demicev: The problem is that even if it's not possible we already allow this option in API, if we remove this now it might be a breaking change for users. As we know from past experience some undocumented options can pass through azure API.

pkg/aks/create.go
```diff
@@ -38,98 +39,123 @@ func CreateCluster(ctx context.Context, cred *Credentials, clusterClient service
...
-		LoadBalancerSku: containerservice.Standard,
+		NetworkPolicy: containerservice.NetworkPolicy(to.String(spec.NetworkPolicy)),
 	}
 
-	if spec.LoadBalancerSKU != nil {
-		networkProfile.LoadBalancerSku = containerservice.LoadBalancerSku(to.String(spec.LoadBalancerSKU))
+	switch containerservice.NetworkPolicy(to.String(spec.NetworkPolicy)) {
+	case containerservice.NetworkPolicyCalico:
+		if containerservice.NetworkPlugin(to.String(spec.NetworkPlugin)) != containerservice.Kubenet {
+			return nil, fmt.Errorf("networkPolicy 'calico' is only supported with networkPlugin 'kubenet'")
```
mjura, line 72: This message is not true, here is compatibility matrix 
```
NetworkPlugin = Kubenet works with NetworkPolicy = ["None", "Calico"]
NetworkPlugin = Azure CNI works with NetworkPolicy = ["None", "Calico", "Azure"]
```

mjura (commented):
It has to be tested, some comments about networkPolicy

mjura (approved):
LGTM

richardcase (approved):
LGTM

PR #137, opened 2023-02-21
Automate operator version bump in rancher

alexander-demicev:
Introduce a GitHub action that uses https://www.updatecli.io/ to automatically bump aks version in rancher when needed. The action should be run manually from GitHub actions tab.


mjura (approved):
LGTM

PR #144, opened 2023-03-03
Assign TenantID to managed cluster identity

MAXxATTAXx:
This is helpful when the Service Principal and the Resource group do not belong in the same Tenant.

Before this change trying to do so would give error showing that the ClientID was not found under TenantID of the Resource group.

### DIFF
--- a/pkg/aks/create.go
+++ b/pkg/aks/create.go
@@ -266,6 +266,13 @@ func createManagedCluster(ctx context.Context, cred *Credentials, workplacesClie
 		managedCluster.APIServerAccessProfile.EnablePrivateCluster = spec.PrivateCluster
 	}
 
+	if cred.TenantID != "" {
+		managedCluster.Identity = &containerservice.ManagedClusterIdentity{
+			TenantID: to.StringPtr(cred.TenantID),
+			Type:     containerservice.ResourceIdentityTypeSystemAssigned,
+		}
+	}
+
 	return managedCluster, nil
 }
 
--- a/pkg/aks/create_test.go
+++ b/pkg/aks/create_test.go
@@ -75,6 +75,7 @@ var _ = Describe("newManagedCluster", func() {
 		cred = &Credentials{
 			ClientID:     "test-client-id",
 			ClientSecret: "test-client-secret",
+			TenantID:     "test-tenant-id",
 		}
 	})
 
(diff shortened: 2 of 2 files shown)


mjura:
@MAXxATTAXx could you provide us instruction, how can we reproduce your scenario? We would like to test it as well.

Thank you so much for opening this PR

mjura (changes requested):
Don't merge, we have to reproduce this problem first

MAXxATTAXx:
In our scenario, we have new environments spawned in exernal Azure tenats to our system. They allow us access on a particular resource_group to deploy and manage an aks cluster. The Service Principal or ClientID belongs to us so the tenantID of the ClientID is different than the tenantID of the resource_group. In azure world it is called Managed Resource Group or Managed Applications.

pkg/aks/create.go
```diff
@@ -236,6 +236,11 @@ func createManagedCluster(ctx context.Context, cred *Credentials, workplacesClie
 		managedCluster.APIServerAccessProfile.EnablePrivateCluster = spec.PrivateCluster
 	}
 
+	managedCluster.Identity = &containerservice.ManagedClusterIdentity{
+		TenantID: &cred.TenantID,
+		Type: "SystemAssigned",
+	}
```
furkatgofurov7, line 242: @MAXxATTAXx hi, thanks for the PR. 

Can you please fix the lint issue?

furkatgofurov7 (changes requested):

pkg/aks/create.go
MAXxATTAXx: Corrected

mjura (changes requested):

pkg/aks/create.go
```diff
@@ -236,6 +236,13 @@ func createManagedCluster(ctx context.Context, cred *Credentials, workplacesClie
 		managedCluster.APIServerAccessProfile.EnablePrivateCluster = spec.PrivateCluster
 	}
 
+	if &cred.TenantID != "" {
```
furkatgofurov7, line 239: @MAXxATTAXx you meant:

```suggestion
	if &cred.TenantID != nil {
```

perhaps, based on the CI [failure](https://drone-pr.rancher.io/rancher/aks-operator/461/1/2) [`pkg/aks/create.go:250:23: invalid operation: &cred.TenantID != "" (mismatched types *string and untyped string)`
]

furkatgofurov7 (changes requested):

pkg/aks/create.go
MAXxATTAXx: THanks for pointing that out. Submitted changes to address this.

furkatgofurov7:
This looks good to me, @MAXxATTAXx can we add some simple tests as well?

furkatgofurov7:
> @furkatgofurov7 Test were added for TenantID to be empty string and when TenantID is provided.

@MAXxATTAXx thanks, lgtm, can you please squash your commits?

@mjura PTAL once again

MAXxATTAXx:
@furkatgofurov7 Commits squashed and rebased to master

mjura (commented):
It looks OK

Could you please add explanation to commit message why this change is needed? Thank you

MAXxATTAXx:
@mjura  I updated the commit message

mjura (approved):
LGTM

alexander-demicev (approved):

furkatgofurov7 (approved):

PR #163, opened 2023-03-25
Enable AKS to use Outbound type of userDefinedRouting

rajan2:
With UDR setup, there is no need to provision loadbalancer for `egress` communication. Egress will be routed to existing route table in the VNET/Subnet. 
This will be useful in the setup where creation of public loadbalancer is prohibited.

### DIFF
--- a/examples/create-example-udr.yaml
+++ b/examples/create-example-udr.yaml
@@ -11,7 +11,7 @@ spec:
   azureCredentialSecret: "REPLACE_WITH_K8S_SECRETS_NAME"
   privateCluster: false,
   linuxAdminUsername: "rancher-user",
-  loadBalancerSku: "standard"
+  loadBalancerSku: ""
   sshPublicKey: "REPLACE_WITH_SSH_PUBLIC_KEY",
   kubernetesVersion: "1.19.9"
   nodePools:
@@ -36,4 +36,4 @@ spec:
     minCount: 1
     maxCount: 6
     availabilityZones: [ "1", "2", "3" ]
-  outboundType: "loadBalancer"
+  outboundType: "userDefinedRouting"
--- a/pkg/aks/create.go
+++ b/pkg/aks/create.go
@@ -64,6 +64,15 @@ func createManagedCluster(ctx context.Context, cred *Credentials, workplacesClie
 
 	networkProfile := &containerservice.NetworkProfile{}
 
+	switch to.String(spec.OutboundType) {
+	case string(containerservice.LoadBalancer):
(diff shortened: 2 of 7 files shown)


rajan2:
@alexander-demicev kindly review

rajan2:
@mjura

pkg/aks/create.go
```diff
@@ -97,7 +106,7 @@ func createManagedCluster(ctx context.Context, cred *Credentials, workplacesClie
 		logrus.Warnf("loadBalancerSKU 'basic' is not supported")
 		networkProfile.LoadBalancerSku = containerservice.Basic
 	case "":
-		networkProfile.LoadBalancerSku = containerservice.Standard
+		networkProfile.LoadBalancerSku = ""
```
furkatgofurov7, line 109: No major comment, but curious why we need to remove this?

rajan2: @furkatgofurov7 ,
Operator creates a standard loadbalancer (egress) by default, by removing this, I will not have egress created by uses pre-defined route table for egress.

mjura: I am not sure about that, but I am going to test it

mjura (commented):
LGTM but it has to be tested

mjura:
Testing in progress ....

mjura:
After some few test runs I see that example-configuration isn't complete:

```
ERRO[1061] error syncing 'cattle-global-data/mjura-outb': handler aks-controller: error failed to create cluster: containerservice.ManagedClustersClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: Code="EmptySubnetError" Message="The input subnet is an empty string." , requeuing 

```

mjura (changes requested):
It still needs more testing and corrections

examples/create-example-udr.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: aks.cattle.io/v1
+kind: AKSClusterConfig
+metadata:
+  name: my-cluster
+spec:
+  resourceLocation: "germanywestcentral"
+  resourceGroup: "my-group"
+  clusterName: "my-cluster"
```
mjura, line 8: it is missing dnsPrefix: "example-string"

mjura (changes requested):
it needs some adjusments

kkaempf:
Will cherry-pick and add fixes

mjura:
We will continue there https://github.com/rancher/aks-operator/pull/189

mjura:
PR https://github.com/rancher/aks-operator/pull/189 was merged

Issue #98, opened 2023-02-03
(SURE-5437) AzureGov Cloud Credentials fail to create/provisioning broken


kkaempf:
#### Describe the bug
When creating a cloud-credential for Azure using the AzureUSGovernmentCloud environment option, the creation will fail. When inspecting the logs in the browser, there is an error referencing 'SubscriptionId not found'. The subscription id has been validated to work as the service principal being used in the credential was created using it.

This error is not present when using standard Azure. It is suspected that AzureGov endpoints are not correctly set internally on a dependent tool.

#### To Reproduce

Create Service Principal in AzureGov using
```
az ad sp create-for-rbac \ -name="<Rancher ServicePrincipal name>" \ --role="Contributor" \ --scopes="/subscriptions/<subscription Id>" * Sign into Rancher UI>Cluster-Management->Cloud-Credentials and create a new Azure Credential
```
Select AzureUSGovernmentCloud for the environment
Plug in generated appId into Client-Id field
Plug in generated password into Client-Secret field
Plug in subscription id into Subscription Id field
Click Create, see non-descriptive error
Open Console and inspect Network and fire command again
Inspect Response to see RESTful error describing the unknown subscription Id

#### Expected Result
Credential Created Successfully

#### Additional context
When bypassing the cloud credential creation using Terraform, the credential itself will fail to work when creating a cluster and repeat the same 'subscription id not found' error.

mjura:
It seems that it is related to https://github.com/rancher/aks-operator/pull/62

PR needs rebase and more refactoring from our site, currently I am working on this

mjura:
Fix https://github.com/rancher/aks-operator/pull/136/

mjura:
Fix was submitted, there is also need to adjust Rancher part for it

kkaempf:
> Fix was submitted, there is also need to adjust Rancher part for it

Which part of Rancher ?!

mjura:
> > Fix was submitted, there is also need to adjust Rancher part for it
> 
> Which part of Rancher ?!

- rancher handler for AKS proxy UI calls https://github.com/rancher/rancher/pull/41001
- we have to also update Cloud Credentials schema for AKS
- and Rancher UI part needs to be also redone

a-blender:
@mjura Merged! Please add a test plan before moving To Test

cpinjani:
@sowmyav27 @anupama2501 Are we having AzureGov access/credentials to validate this issue ?
cc: @prabalsharma

[...]

PR #171, opened 2023-04-08
Add credentials to buildUpstreamClusterState func

mjura:
Issue: https://github.com/rancher/aks-operator/issues/98


mjura:
It was fixed already in https://github.com/rancher/aks-operator/pull/174

PR #174, opened 2023-04-11
🐛  Fix aks credentials passing to buildUpstreamClusterState() func

furkatgofurov7:
Possibly fixes: https://github.com/rancher/aks-operator/issues/98

Signed-off by: Furkat Gofurov (furkat.gofurov@suse.com)

PR #175, opened 2023-04-13
Refactor: move CODEOWNERS to the root of the repo

mbologna:
Moving CODEOWNERS to the root (uniforming with Rancher repos)


furkatgofurov7 (approved):
CI check will be failing in all PRs in the repo until we either merge https://github.com/rancher/aks-operator/pull/174 or https://github.com/rancher/aks-operator/pull/171 (both should fix the same, but with slightly different approach and coverage).

PR #176, opened 2023-04-17
Update k8s version in e2e tests

alexander-demicev:
Update k8s version because 1.24.6 was deprecated by Microsoft.

### DIFF
--- a/test/e2e/templates/basic-cluster.yaml
+++ b/test/e2e/templates/basic-cluster.yaml
@@ -7,7 +7,7 @@ spec:
   clusterName: cluster
   dnsPrefix: basic-cluster-dns
   imported: false
-  kubernetesVersion: 1.24.6
+  kubernetesVersion: 1.25.5
   linuxAdminUsername: azureuser
   loadBalancerSku: Standard
   networkPlugin: kubenet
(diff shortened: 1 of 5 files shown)


test/e2e/templates/basic-cluster.yaml
```diff
@@ -7,7 +7,7 @@ spec:
   clusterName: cluster
   dnsPrefix: basic-cluster-dns
   imported: false
-  kubernetesVersion: 1.24.6
+  kubernetesVersion: 1.25.5
```
furkatgofurov7, line 10: I think we need to update the kind image version we use to set up the kind cluster to make it aligned since it is using the same versions: https://github.com/rancher/aks-operator/blob/ee6e4a03ebf58cb752ff2bcb81f5f65f82f601e3/Makefile#L10 and https://github.com/rancher/aks-operator/blob/9cef3583b32390099a7a7d3a1477cb05b445d59b/scripts/setup-kind-cluster.sh#L3

wdyt?

furkatgofurov7 (commented):
Can we update the PR description with the reference to the deprecation info (I could not find it myself, so it would be helpful)?

test/e2e/templates/basic-cluster.yaml
alexander-demicev: These versions can be different but it's a good idea to bump kind too. Thanks for spotting it.

alexander-demicev: done

furkatgofurov7 (approved):

Issue #169, opened 2023-04-04
(SURE-6076) AKS TF-provisioned private cluster import is failing


kkaempf:
### Issue description:
AKS private cluster import fails with the error: "cannot detect cluster [xxxx] upstream DNS prefix"

### Business impact:
Not being able to manage AKS clusters through Rancher.

### Troubleshooting steps:

Import a private AKS cluster using the Rancher new AKS cluster Registration command=-> KO (dns error msg)
Create a new private AKS cluster using Rancher UI, and used 'az command-invoke' to register the AKS cluster in Rancher> KO (same error msg)
Create a new private AKS cluster using Azure portal, used Rancher UI to import the cluster without success. Used 'az command-invoke' from az cli, to import the cluster, with no success.
Create a new private AKS cluster using TF, used the Rancher UI to import the cluster => KO 
Verified that Rancher is reachable (curl to Rancher_Server URL and name resolvable), from within a pod of a newly TF-provisioned private AKS cluster
Set Rancher in Debug mode (showed the error message, but no more evidences)

### Repro steps:

Provision a private AKS cluster TF, using  azure_rm_provider, using a custom vnet and custom private DNS zone
Import the cluster into Rancher from the UI

### Workaround:
Is a workaround available and implemented? yes
What is the workaround: Import the AKS clusters as custom clusters

### Actual behavior:
AKS private clusters cannot be imported into Rancher.

### Expected behavior:
AKS private clusters with custom networking (existing vnet) and custom private DNS zone can be imported into Rancher.

### Files, logs, traces:
Attached is the Rancher debug log, traces of DNS and AKS reachability from on-prem, and the Rancher reachability test from Azure.

### Additional notes:

Customer's Rancher server is hosted in an on-premises environment, with an on-prem network.
AKS clusters are private, and deployed in private existing networks (vnet) and custom DNS zone.  On-premises network is connected to Azure networks via ExpressRoute
AKS clusters can be directly managed from the customer's on-prem environment using kubectl with the kubeconfig credentials downloaded via az cli.

mjura:
I have got requested configuration files, I am investigating it once again

mjura:
Fix was submitted https://github.com/rancher/aks-operator/pull/178

mjura:
Command which I used to build AKS cluster:

```
az aks create -g mjura-eastus -n mjura-outbound -l eastus \
      --node-count 2 --network-plugin kubenet --network-policy calico \
[...]

PR #178, opened 2023-04-24
Skip updating dnsPrefix for private DNS zones

mjura:
For imported private clusters with private DNS zones we can skip setting dnsPrefix, because it will be nil value. DNSPrefix can be updated for existing clusters.

Issue: https://github.com/rancher/aks-operator/issues/169

### DIFF
--- a/controller/aks-cluster-config-handler.go
+++ b/controller/aks-cluster-config-handler.go
@@ -567,10 +567,9 @@ func (h *Handler) buildUpstreamClusterState(ctx context.Context, credentials *ak
 	upstreamSpec.KubernetesVersion = clusterState.KubernetesVersion
 
 	// set DNS prefix
-	if clusterState.DNSPrefix == nil {
-		return nil, fmt.Errorf("cannot detect cluster [%s] upstream DNS prefix", spec.ClusterName)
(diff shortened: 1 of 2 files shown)


controller/aks-cluster-config-handler.go
```diff
@@ -567,10 +567,9 @@ func (h *Handler) buildUpstreamClusterState(ctx context.Context, credentials *ak
 	upstreamSpec.KubernetesVersion = clusterState.KubernetesVersion
 
 	// set DNS prefix
-	if clusterState.DNSPrefix == nil {
-		return nil, fmt.Errorf("cannot detect cluster [%s] upstream DNS prefix", spec.ClusterName)
+	if clusterState.DNSPrefix != nil {
+		upstreamSpec.DNSPrefix = clusterState.DNSPrefix
```
furkatgofurov7, line 571: Not a blocker but a clarification question: what happens in case when `existing cluster DNSPrefix == nil`

mjura: We would like to make possible only import AKS clusters with privateDNS zones.

furkatgofurov7 (commented):
LGTM

furkatgofurov7 (approved):

PR #182, opened 2023-04-27
Prevent rancher from downgrading chart version

alexander-demicev:
If the chart version is set to some unrealistically high version like 900, rancher will not reconcile it. Currently, when aks controller starts returning errors rancher will try to fix the chart:

```
mac:aks-operator ademicev$ kubectl get deployment aks-config-operator -n cattle-system -o yaml | grep image:
        image: ghcr.io/rancher/aks-operator:v0.0.0-fbfbe49
mac:aks-operator ademicev$ kubectl get deployment aks-config-operator -n cattle-system -o yaml | grep image:
        image: rancher/aks-operator:v1.1.0
```
changes in the PR will prevent this from happening


furkatgofurov7 (commented):
LGTM

mjura (approved):
awesome :+1:

richardcase (approved):

PR #192, opened 2023-05-08
🌱 Enable more linters in aks operator

furkatgofurov7:
This PR:

- Enables more linters in the .golangci.yaml to have a better aligned package import aliases
- Fixes all lint errors as a consequence of newly added linters

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,19 +1,33 @@
 run:
   timeout: 5m
+  go: "1.19"
   skip-files:
     - "zz_generated_*"
   skip-dirs:
     - pkg/generated
   tests: false
+  allow-parallel-runners: true
+
 output:
   format: github-actions
+
 linters:
+  disable-all: true
   enable:
-    - revive # replacement for golint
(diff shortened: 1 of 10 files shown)


.golangci.yml
```diff
@@ -1,19 +1,33 @@
...
   skip-dirs:
     - pkg/generated
   tests: false
+  allow-parallel-runners: true
+
 output:
   format: github-actions
+
 linters:
+  disable-all: true
```
richardcase, line 15: As a personal preference i much prefer enable all and then explicitly disabling linters.

richardcase (approved):

.golangci.yml
furkatgofurov7: While I agree with your point, I found `enable-all: true ` be disruptive at times, because under the hood it means: whenever a new linter is added or upgraded upstream, it will try to enable it thus making the probability of build failures high and somewhat uncontrollable, whereas `disable-all: true` and providing the specific linters we want to use to `enable:` list is much more controlled.

PR #232, opened 2023-06-26
Always use CI chart for E2E tests

alexander-demicev:
Setting `CATTLE_SKIP_HOSTED_CLUSTER_CHART_INSTALLATION` will allow us to always be sure that we are running an operator chart built for CI.
Later when rancher 2.7.5 is released we can use it instead of RC.


mjura (approved):
LGTM

PR #253, opened 2023-08-16
[v2.8] Bump go version to 1.20

mjura:


salasberryfin (dismissed):

salasberryfin:
I see the new release branches are not yet protected (e.g. we can merge with no reviews). I think we should align these branches' configuration with master's.

mjura:
> I see the new release branches are not yet protected (e.g. we can merge with no reviews). I think we should align these branches' configuration with master's.

Thank you, I have updated branches settings with following regexp `[master,release-]*`

salasberryfin (approved):

Issue #214, opened 2023-06-09
Remove dependency on rancher-machine


richardcase:
It would be good to remove our dependency on `rancher-machine`. We currently use it in 1 place to get the tenant id:

https://github.com/rancher/aks-operator/blob/master/pkg/aks/client.go#L140

mjura:
This can be closed

Issue #287, opened 2023-09-28
Migrate to the new Azure SDK modules


mjura:
"github.com/Azure/azure-sdk-for-go/services/subscription/mgmt/*" is deprecated: Please note, this package has been deprecated. A replacement package is available [github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/subscription/armsubscription](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/subscription/armsubscription). We strongly encourage you to upgrade to continue receiving updates. See [Migration Guide](https://aka.ms/azsdk/golang/t2/migration) for guidance on upgrading. Refer to our [deprecation policy](https://azure.github.io/azure-sdk/policies_support.html) for more details.

Known issues:
- [ ] https://github.com/rancher/aks-operator/issues/248

mjura:
This should also include problem with removing deprecated NetworkDockerBridgeCIDR from spec field https://github.com/rancher/aks-operator/pull/338

mjura:
it was already done in https://github.com/rancher/aks-operator/issues/132

PR #308, opened 2023-11-23
Remove dependency on rancher-machine module 

mjura:

We have to drop `rancher-machine` module dependency, because it was deprecated.

Add FindTenantID and NewSubscriptionsClient functions.

FindTenantID function is used to find the tenant ID for the subscription ID. It will send an unauthenticated request to the Azure Resource Manager endpoint to get the tenant ID from the WWW-Authenticate header.

Issue #https://github.com/rancher/aks-operator/issues/214


pkg/aks/client.go
```diff
@@ -3,16 +3,18 @@ package aks
 import (
 	"context"
 	"fmt"
+	"net/http"
+	"regexp"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2020-11-01/containerservice"
+	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2019-11-01/subscriptions"
```
Danil-Grigorev, line 11: Could the version of the azure sdk be unified here? 2020-11-01 for both.

mjura: Unfortunately this version doesn't exist, we are using already `2019-11-01` next is completely new Azure SDK and migration will be done in another  card https://github.com/rancher/aks-operator/issues/287

Danil-Grigorev (approved):

furkatgofurov7 (approved):

PR #309, opened 2023-11-23
[2.8] Remove dependency on rancher-machine module

mjura:

We have to drop rancher-machine module dependency, because it was deprecated.

Add FindTenantID and NewSubscriptionsClient functions.

FindTenantID function is used to find the tenant ID for the subscription ID. It will send an unauthenticated request to the Azure Resource Manager endpoint to get the tenant ID from the WWW-Authenticate header.

Issue https://github.com/rancher/aks-operator/issues/214


furkatgofurov7:
@mjura this probably needs a rebase? https://github.com/rancher/aks-operator/pull/311 is already merged and should fix scan problems in CI

PR #335, opened 2023-12-07
🌱 Add azure-sdk-for-go dependency to dependabot ignore list in all branches

furkatgofurov7:
Adds `azure-sdk-for-go` dependency to dependabot ignore list in all branches. We are intending to track bumping of that dependency in https://github.com/rancher/aks-operator/issues/287. Once #287 is completed, it can be removed from ignore list

- [ ] includes documentation
- [ ] adds unit tests
- [ ] backport needed


mjura (approved):
let's do this

PR #407, opened 2024-02-20
[v2.9] Revert github.com/rancher/rancher/pkg/apis version

mjura:

Issue #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests
- [ ] backport needed

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/rancher-sandbox/ele-testhelpers v0.0.0-20221213084338-a8ffdd2b87e3
 	github.com/rancher/lasso v0.0.0-20230830164424-d684fdeb6f29
-	github.com/rancher/rancher/pkg/apis v0.0.0-20240126142034-676c3eb3dfa5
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -19,7 +19,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/rancher-sandbox/ele-testhelpers v0.0.0-20221213084338-a8ffdd2b87e3
 	github.com/rancher/lasso v0.0.0-20230830164424-d684fdeb6f29
-	github.com/rancher/rancher/pkg/apis v0.0.0-20240126142034-676c3eb3dfa5
+	github.com/rancher/rancher/pkg/apis v0.0.0-20230915232223-a9ea4ce4a5ba
```
salasberryfin, line 22: This is a downgrade compared to what we have in `v2.8`, [here](https://github.com/rancher/aks-operator/blob/8e23ef9164d73a0b543a28ca13ea8dc8dfc3fdc9/go.mod#L19C45-L19C72). Perhaps we can use the same version we have there?

mjura: yes, it is because Rancher v2.8 was already migrated to wrangler v2.0 which requires new `github.com/rancher/rancher/pkg/apis`

salasberryfin (approved):

PR #568, opened 2024-06-25
[v2.7] Bump Go version to 1.21.11

yiannistri:

Bumped Go version to 1.21.11.

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests
- [ ] backport needed


yiannistri:
It turns out it's a bit risky to bump the Go version to 1.21 while rancher 2.7 remains on 1.20. Such a change would cause rancher's Go version to get bumped as well and that should be a deliberate decision not a side-effect of updating a dependency. Therefore closing this PR.

Issue #592, opened 2024-07-11
Unable to remove an authorized IP range


valaparthvi:
Rancher: v2.9-3c4ccdc5bc9fde3510089153b5ad58fdbe604880-head
Dashboard: master https://github.com/rancher/dashboard/commit/00e9eb2a64b5bde237fd9b9e03a03b460e8d7307

Adding a new IP range works as expected, but removing an IP range does not work, from UI, all the buttons are active and even the operator logs indicate that an update is in progress, but when the update finishes, there appears to be no change.

Steps to reproduce:
1. Create an AKS cluster with two authorized IP ranges.
2. Once the cluster is active, update it to remove one of the ranges OR uncheck `Authorized IP Ranges`.
3. Monitor deploy/aks-config-operator logs to ensure the update is in progress.
4. Once the update finishes, check the config.

The removed IP range is still present in both AKSConfig and UpstreamSpec.

_Notes:_
1. The cluster does not go into `Updating` state, is this expected?
2. Editing an existing IP range, adds a new one.

Ref: https://github.com/rancher/dashboard/issues/6879#issuecomment-2222294833

**PR's**
- [x] https://github.com/rancher/aks-operator/pull/595
- [x] https://github.com/rancher/aks-operator/pull/594

valaparthvi:
@mjura @yiannistri Can this get some love?

valaparthvi:
Fixed on Rancher v2.9-6e9cf233a87796ee9a96e73c096a82101fc9c2ff-head.

valaparthvi:
cc: @mjura @yiannistri

valaparthvi:
Tested this again on Rancher (v2.9-0ab20e8fe6275887c52b97e94e15365f33b8afab-head); Dashboard(release-2.9 https://github.com/rancher/dashboard/commit/6e45e2fabb473335ca8975560b57d1657554ff27), aks-operator(rancher-aks-operator:104.1.0+up1.9.1-rc.4).

FIXED.

PR #594, opened 2024-07-19
[v2.9] Fix logic for Authorized IP ranges

mjura:
Fix logic for updating Authorized IP ranges:
- all settings should be overwrite for AKS clusters provisioned by Rancher
- for imported clusters options should be merged with downstream cluster

Issue: https://github.com/rancher/aks-operator/issues/592

Issue #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests
- [ ] backport needed

### DIFF
--- a/pkg/aks/update.go
+++ b/pkg/aks/update.go
@@ -48,7 +48,7 @@ func UpdateCluster(ctx context.Context, cred *Credentials, clusterClient service
 		ctx,
 		spec.ResourceGroup,
 		spec.ClusterName,
-		updateCluster(*desiredCluster, actualCluster.ManagedCluster),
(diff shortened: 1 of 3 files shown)


pkg/aks/update.go
```diff
@@ -108,14 +108,23 @@ func updateCluster(desiredCluster armcontainerservice.ManagedCluster, actualClus
 		}
 	}
 
+	// If the cluster is imported, we may not have the authorized IP ranges set on the spec.
```
furkatgofurov7, line 111: Q: is there a case where  authorized IP ranges are set on the spec and if so what happens in that case?

mjura: For imported clusters it is not possible, always this setting will be overwritten with downstream cluster

yiannistri (approved):

Issue #591, opened 2024-07-10
SURE-8268


kkaempf:
See SURE-8268 for details

All operators are affected

**PR's**
AKS
- https://github.com/rancher/aks-operator/pull/605
- https://github.com/rancher/aks-operator/pull/604

EKS
- https://github.com/rancher/eks-operator/pull/710
- https://github.com/rancher/eks-operator/pull/709

GKE
- https://github.com/rancher/gke-operator/pull/570
- https://github.com/rancher/gke-operator/pull/571

PR #604, opened 2024-07-23
Improve securityContext for operator Deployment

mjura:
1. Restrict container from acquiring additional privileges (`securityContext.allowPrivilegeEscalation`)
2. Mount container's root filesystem as read only (`securityContext.readOnlyRootFilesystem`)
3. Ensure that container won't be started as privileged container (`securityContext.privileged`)

Issue: https://github.com/rancher/aks-operator/issues/591

Issue #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests
- [ ] backport needed

### DIFF
--- a/charts/aks-operator/templates/deployment.yaml
+++ b/charts/aks-operator/templates/deployment.yaml
@@ -39,6 +39,13 @@ spec:
           value: {{ .Values.httpsProxy }}
         - name: NO_PROXY
           value: {{ .Values.noProxy }}
+        securityContext:
+          allowPrivilegeEscalation: false
+          readOnlyRootFilesystem: true
+          privileged: false
+          capabilities:
+            drop:
+            - ALL
 {{- if .Values.additionalTrustedCAs }}
         # aks-operator mounts the additional CAs in two places:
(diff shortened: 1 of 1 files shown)


Danil-Grigorev (dismissed):

yiannistri (changes requested):
I believe these settings should be under `.spec.containers[n].securityContext` rather than `spec.securityContext`. See https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod and https://kubernetes.io/docs/tasks/configure-pod-container/security-context/

mjura:
> I believe these settings should be under `.spec.template.spec.containers[n].securityContext` rather than `spec.template.spec.securityContext`. See https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod and https://kubernetes.io/docs/tasks/configure-pod-container/security-context/

you are right, I discovered it after testing. thank you for pointing this

yiannistri (approved):
:shield: :ok_hand:

Issue #665, opened 2024-09-10
AKS: Unexpected error in logs while recording failure message


cpinjani:
**Rancher version:**
Rancher - v2.9-65df012014c22ba29006fca8c97aa6b8c2ba1105-head
aks-operator:v1.9.2-rc.1

Cluster Type: Downstream AKS cluster

**Describe the bug:**
Unexpected error in logs while recording failure message
See also: https://github.com/rancher/eks-operator/issues/596

**Logs:**
```
time="2024-09-10T12:50:01Z" level=info msg="Create Azure Log Analytics Workspace \"cpinjani-aks4-EUS\" on Resource Group \"cpinjani-aks4\""
time="2024-09-10T12:53:38Z" level=error msg="Error recording akscc [ (id: )] failure message: resource name may not be empty"
```

```
time="2024-09-10T10:22:21Z" level=info msg="Importing config for cluster [cpinjani-imp (id: c-sflm4)]"
time="2024-09-10T10:22:22Z" level=info msg="Importing config for cluster [cpinjani-imp (id: c-sflm4)]"
time="2024-09-10T10:22:23Z" level=error msg="Error recording akscc [ (id: )] failure message: resource name may not be empty"
time="2024-09-10T10:22:25Z" level=error msg="Error recording akscc [ (id: )] failure message: resource name may not be empty"
time="2024-09-10T10:22:26Z" level=info msg="Checking configuration for cluster [cpinjani-imp (id: c-sflm4)]"
```

yiannistri:
The error originates [here](https://github.com/rancher/aks-operator/blob/bf1310f116a0001d59f74c240b30a982fdca8273/controller/aks-cluster-config-handler.go#L187) when a conflict occurs. When that happens, the `config` object returned is non-nil but empty which is why the log statement appears to be missing fields.

thehejik:
### Test report
The problem seems to be fixed in latest available version of AKS operator `rancher-aks-operator:104.3.0+up1.9.3-rc.1` on rancher `v2.9-43c67f5d91757836c8fa732be35c6171d89759c1-head`.

#### Reproducer
• provision AKS from rancher with default values
• via Edit Config in Rancher add a NP to the AKS cluster with unsupported value for Max Pods per node = 1  (minimal allowed is 10)
• open and watch aks operator logs
• wait for RESPONSE 400: `400 Bad Request\nERROR CODE: InsufficientMaxPods`
[...]

PR #668, opened 2024-09-20
fix: Retry operation on conflict

yiannistri:

Occasionally when a conflict occurs during the reconciliation process, the `config` object that is returned is non-nil but empty, causing logs to be missing fields. 

Issue #665 

To address the conflict, the [RetryOnConflict function](https://pkg.go.dev/k8s.io/client-go/util/retry#RetryOnConflict) was used to query the resource and attempt the operation again, if the error returned is a conflict error.

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests
- [ ] backport needed

### DIFF
--- a/controller/aks-cluster-config-handler.go
+++ b/controller/aks-cluster-config-handler.go
@@ -19,6 +19,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/util/retry"
 
 	"github.com/rancher/aks-operator/pkg/aks"
 	"github.com/rancher/aks-operator/pkg/aks/services"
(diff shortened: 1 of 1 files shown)


controller/aks-cluster-config-handler.go
```diff
@@ -184,7 +185,25 @@ func (h *Handler) recordError(onChange func(key string, config *aksv1.AKSCluster
...
+				return nil
+			}
+
+			conf, err := h.aksCC.Get(config.Namespace, config.Name, metav1.GetOptions{})
+			if err != nil {
+				return err
+			}
+
+			conf, err = onChange(key, conf)
+			if err == nil {
```
mjura, line 201: I think we can update `config` regardless of whether `err `is nil to reflect any changes made by `onChange`, even in error scenarios.

```suggestion
			config = conf
```

controller/aks-cluster-config-handler.go
```diff
@@ -184,7 +185,25 @@ func (h *Handler) recordError(onChange func(key string, config *aksv1.AKSCluster
 	return func(key string, config *aksv1.AKSClusterConfig) (*aksv1.AKSClusterConfig, error) {
 		var err error
 		var message string
-		config, err = onChange(key, config)
+
+		err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			// AKS config is being deleted
+			if config == nil {
```
mjura, line 191: We are duplicating this check, look on line#207

mjura (dismissed):
it looks ok, good job

controller/aks-cluster-config-handler.go
```diff
@@ -184,7 +185,25 @@ func (h *Handler) recordError(onChange func(key string, config *aksv1.AKSCluster
...
+				return nil
+			}
+
+			conf, err := h.aksCC.Get(config.Namespace, config.Name, metav1.GetOptions{})
```
mjura, line 195: if conf == nil {
    return fmt.Errorf("EKSClusterConfig %s/%s not found", config.Namespace, config.Name)
}

controller/aks-cluster-config-handler.go
yiannistri: The reason for this check is to avoid referencing a nil object when we query for a fresh copy. This is possible if the config has been deleted.

controller/aks-cluster-config-handler.go
yiannistri: Unfortunately when a conflict occurs, the config object returned from `onChange` is empty/non-nil which causes the original bug. That is the sole reason for checking if `err` is nil.

salasberryfin (dismissed):

mjura (approved):

mjura:
Please backport this change to v2.9

Issue #667, opened 2024-09-16
Changing multiple fields of cluster while update is not working


cpinjani:
**Rancher version:**
Rancher - v2.9-f1b43d2568d7c53c3adf45d9ffd74a04ea65fc22-head
aks-operator:v1.9.2-rc.2

Cluster Type: Downstream AKS cluster

**Describe the bug:**
- Provision cluster with multiple Nodepools and wait for it to be Active
- Edit the cluster: delete any Nodepool and enable HTTP Application routing, click Save.

Spec is updated with changes applied.
Only the change to routing gets applied, and Nodepool remains intact after the update.
```
aksConfig:
    authBaseUrl: https://login.microsoftonline.com/
    authorizedIpRanges: null
    azureCredentialSecret: cattle-global-data:cc-j7sg7
    baseUrl: https://management.azure.com/
    clusterName: cpinjani-aks
    dnsPrefix: cpinjani-aks
    dnsServiceIp: 10.0.0.10
    dockerBridgeCidr: null
    httpApplicationRouting: true
    imported: false
    kubernetesVersion: 1.29.0
    linuxAdminUsername: azureuser
    loadBalancerSku: Standard
    logAnalyticsWorkspaceGroup: null
    logAnalyticsWorkspaceName: null
    managedIdentity: null
    monitoring: null
    networkPlugin: kubenet
    networkPolicy: null
    nodePools:
      - availabilityZones:
          - '1'
          - '2'
          - '3'
        count: 1
        maxPods: 110
        maxSurge: '1'
        mode: System
        name: np1
        orchestratorVersion: 1.29.0
        osDiskSizeGB: 128
        osDiskType: Managed
        osType: Linux
        vmSize: Standard_DS2_v2
    outboundType: loadBalancer
    podCidr: 10.244.0.0/16
    privateCluster: false
    privateDnsZone: null
    resourceGroup: cpinjani-aks
    resourceLocation: eastus
    serviceCidr: 10.0.0.0/16
    subnet: null
    tags:
      Account Type: group
    userAssignedIdentity: null
    virtualNetwork: null
    virtualNetworkResourceGroup: null
```

**Logs:**

2.9:
```
time="2024-09-16T11:03:50Z" level=info msg="Checking configuration for cluster [cpinjani-aks (id: c-jm2vn)]"
time="2024-09-16T11:03:50Z" level=info msg="Updating HTTP application routing to true for cluster [cpinjani-aks (id: c-jm2vn)]"
[...]

PR #703, opened 2024-10-28
Correct code flow of node pools

vatsalparekh:

Issue #667 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests
- [ ] backport needed

### DIFF
--- a/controller/aks-cluster-config-handler.go
+++ b/controller/aks-cluster-config-handler.go
@@ -843,18 +843,19 @@ func (h *Handler) updateUpstreamClusterState(ctx context.Context, config *aksv1.
 			clusterSpecCopy.ResourceGroup = config.Spec.ResourceGroup
 			clusterSpecCopy.ResourceLocation = config.Spec.ResourceLocation
 			clusterSpecCopy.ClusterName = config.Spec.ClusterName
-		}
-		clusterSpecCopy.NodePools = make([]aksv1.AKSNodePool, 0, len(config.Spec.NodePools))
-		for _, n := range config.Spec.NodePools {
-			if _, ok := upstreamNodePools[*n.Name]; ok {
-				clusterSpecCopy.NodePools = append(clusterSpecCopy.NodePools, n)
+
+			clusterSpecCopy.NodePools = make([]aksv1.AKSNodePool, 0, len(config.Spec.NodePools))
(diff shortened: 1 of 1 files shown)


controller/aks-cluster-config-handler.go
```diff
@@ -843,18 +843,19 @@ func (h *Handler) updateUpstreamClusterState(ctx context.Context, config *aksv1.
 			clusterSpecCopy.ResourceGroup = config.Spec.ResourceGroup
 			clusterSpecCopy.ResourceLocation = config.Spec.ResourceLocation
 			clusterSpecCopy.ClusterName = config.Spec.ClusterName
-		}
-		clusterSpecCopy.NodePools = make([]aksv1.AKSNodePool, 0, len(config.Spec.NodePools))
-		for _, n := range config.Spec.NodePools {
-			if _, ok := upstreamNodePools[*n.Name]; ok {
-				clusterSpecCopy.NodePools = append(clusterSpecCopy.NodePools, n)
```
mjura, line 850: Why are you changing logic for provisioned clusters by Rancher? In results of this change non-imported clusters won't be updated.

mjura (changes requested):

controller/aks-cluster-config-handler.go
vatsalparekh: The loop iterating over config.NodePools checks if that NodePool is present in upstreamNodePools, which means that no new pool can be added or deleted. This logic seems confusing. Changing this only for Imported cluster seems logical

mjura: Updates for NodePools need to work for both rancher managed clusters and imported one. Your change will break this and only imported cluster will be updated.

mjura (changes requested):

yiannistri:
After testing this a couple of times, the behaviour described in https://github.com/rancher/aks-operator/issues/667 is still present:

1. Created a cluster with 1 system node pool and 3 additional user node pools, each with 1 node.
2. Once active, I removed one of the user node pools while also enabling the HTTP application routing.
3. After a few minutes, the node pool was removed but the HTTP application routing remains disabled.

yiannistri:
As there have been no updates in this PR for some time, I think  it's best to close it for now.

Issue #748, opened 2024-12-11
SLSA 3 compliant release process


yiannistri:
Update the release process:
- [x] Ensure the build is [self-contained](https://github.com/rancher/rancher/pull/46633/commits/ec46c02a41227e63b533bb8e130284047a5ccbb5)
- [x] Add/update Makefile targets that build/push images as per guidance
- [x] Update GHA release workflow

yiannistri:
Putting this on-hold until https://github.com/rancherlabs/eio/issues/2883 is resolved.

PR #749, opened 2024-12-11
ci: Make release SLSA compliant

yiannistri:

This PR changes the build process to be self-contained: it no longer requires the application to be build outside the container and relying on go version in the CI runner, instead the build is done in a build layer.

It also updates the release workflow to additionally push images to Prime Registry.

Finally it uses [cross-compilation helpers](https://github.com/tonistiigi/xx?tab=readme-ov-file#go--cgo) to invoke the Go compiler with the correct env vars for the target platform.

Issue #748 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests
- [ ] backport needed

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -1,17 +1,26 @@
-FROM registry.suse.com/bci/bci-base:15.6 AS builder
+FROM registry.suse.com/bci/bci-base:15.6 AS base
 RUN sed -i 's/^CREATE_MAIL_SPOOL=yes/CREATE_MAIL_SPOOL=no/' /etc/default/useradd
 RUN useradd --uid 1007 aks-operator
 
+FROM registry.suse.com/bci/golang:1.23 AS builder
+WORKDIR /app
+COPY go.mod go.sum ./
+RUN go mod download && go mod verify
+COPY ./controller ./controller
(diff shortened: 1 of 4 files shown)


pjbgf (approved):
LGTM

package/Dockerfile
```diff
@@ -1,17 +1,26 @@
...
 RUN useradd --uid 1007 aks-operator
 
+FROM registry.suse.com/bci/golang:1.23 AS builder
+WORKDIR /app
+COPY go.mod go.sum ./
+RUN go mod download && go mod verify
+COPY ./controller ./controller
+COPY ./pkg ./pkg
+COPY ./main.go ./main.go
+RUN CGO_ENABLED=0 GOOS=linux go build -o /aks-operator
```
pjbgf, line 12: Due to the cross compilation, it may be sensible to [verify](https://github.com/rancher/cis-operator/blob/main/package/Dockerfile#L60) the generated binary is valid for the target architecture.

yiannistri: Good idea :+1:  I've updated the PR to use cross-compilation helpers.

mjura (approved):

Issue #802, opened 2025-01-08
[aks-operator] VEX related improvements for the build process


macedogm:
The VEX Hub initiative in [rancher/vexhub](https://github.com/rancher/vexhub) is now a mature project in Rancher where we generate VEX reports for known false-positive CVEs in our projects (images and binaries).

In order for VEX to fully work and for security scanners (e.g., Trivy) to correctly match a VEX entry in our VEX Hub with a Go binary being scanned, the Go binary must have its full package path inside of it. Example:

```
> go version -m aks-operator | head -n 3
bin/aks-operator: go1.23.4
	path	github.com/rancher/aks-operator
	mod	github.com/rancher/aks-operator	(devel)	
```

When a Go binary is compiled by specifying directly the file as `go build main.go` as opposed to `go build .`, the binary won't contain the package path.

```
> go version -m aks-operator | head -n 2
main: go1.23.4
	path	command-line-arguments
```

In the above case, the security scanner won't be able to match the binary with its respective VEX entry.

Further automation that we developed internally also relies on identifying the commit ID from where the binary was built. Example:

```
> go version -m aks-operator
(...)
	build	-ldflags="-X github.com/rancher/aks-operator/pkg/version.GitCommit=05494418fb9e1090521028b522745c8e575be8bd             -X github.com/rancher/aks-operator/pkg/version.Version=v0.0.0-0549441"
(...)
	build	vcs.revision=05494418fb9e1090521028b522745c8e575be8bd
(...)
```

Some branches of `aks-operator` have the needed package path, but lack the commit ID. We'll propose some PRs to help improve the build process and add the needed information for Rancher's VEX Hub project. None of those PRs are expected to affect the code's behavior and features, as they are only metadata information.

kkaempf:
closing as fixed

Issue #726, opened 2024-11-25
Log build commit SHA when operator starts


yiannistri:
When starting the operator, it would be helpful to log the version (ideally the commit SHA used to build the binary) for debugging reasons. This way, users can be confident of which version (and from which commit) the binary originates from.

thehejik:
Works as expected, tested on rancher `v2.10-1787d0b85b6acbfbe1468a21458c2d8a13cca03e-head` with custom built aks-operator from `main` branch.

After execution the version and commit are reported:
```bash
./aks-operator 
INFO[0000] Starting aks-operator (version: 2025.01.10, commit: f6a3406d4dc5b86d49e1ca84730b09f293c9dca5) 
INFO[0000] Starting aks.cattle.io/v1, Kind=AKSClusterConfig controller 
INFO[0000] Starting /v1, Kind=Secret controller
```

PR #803, opened 2025-01-08
VEX related improvements in the build process

macedogm:

Issue https://github.com/rancher/aks-operator/issues/802

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests
- [x] backport needed

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -19,8 +19,14 @@ COPY ./main.go ./main.go
 COPY --from=xx / /
 
 ARG TARGETPLATFORM
+ARG COMMIT
+ARG VERSION
 ENV CGO_ENABLED=0
-RUN xx-go build -o /aks-operator && xx-verify /aks-operator
+RUN xx-go build -ldflags \
+           "-X github.com/rancher/aks-operator/pkg/version.GitCommit=${COMMIT} \
+            -X github.com/rancher/aks-operator/pkg/version.Version=${VERSION}" \
+       -o /aks-operator && \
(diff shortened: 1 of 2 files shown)


mjura (approved):
LGTM

package/Dockerfile
```diff
@@ -19,8 +19,14 @@ COPY ./main.go ./main.go
 COPY --from=xx / /
 
 ARG TARGETPLATFORM
+ARG COMMIT
+ARG VERSION
 ENV CGO_ENABLED=0
-RUN xx-go build -o /aks-operator && xx-verify /aks-operator
+RUN xx-go build -ldflags \
+           "-X github.com/rancher/aks-operator/pkg/version.GitCommit=${COMMIT} \
```
yiannistri, line 26: Don't we need to include a `version` package with `GitCommit` and `Version` variables for this to work? From a quick look, I didn't find such a package.

macedogm: In case the build was happening with Go having access to the `.git` dir, then `go build` would add the metadata info directly in `vcs.revision` and other related fields. Given that we are building inside the container and only copying some of the Go related files, `go build` cannot extract those metadata. That's why I proposed to add them.

yiannistri: Ok cool. I was thinking about doing this exact thing for https://github.com/rancher/aks-operator/issues/726 but I will make a follow-up PR once this PR gets merged.

macedogm: Nice 👍🏻

Issue #751, opened 2024-12-12
API does not throw error while adding nodepool with AZ 


cpinjani:
**Setup**
Rancher version: `v2.10-c4944b3ca0b39c474121df9e7ef31f7aff4e4f7a-head`
AKS Operator version: `v1.10.1-rc.2`

**Description**
Via UI, it is not possible to add nodepool with Availability Zone for a cluster created in region with no support for AZ. Updating this via API does not raise any error. It simply reverts the changes to AKSConfig after a few minutes.
The API should raise an error for the same.

**Steps to reproduce:**
- Create a cluster in region without AZ support Eg. WestUS
- Add nodepool via API and mention availability zones

**Expected results:**
The API must throw an error while adding nodepool with AZ in region where AZ are not supported.

**Logs**
```
time="2024-12-12T08:17:17Z" level=info msg="Checking configuration for cluster [cpinjani-akswest (id: c-4zdpb)]"
time="2024-12-12T08:17:18Z" level=info msg="Adding node pool [pool11] for cluster [cpinjani-akswest (id: c-4zdpb)]"
time="2024-12-12T08:17:24Z" level=info msg="Waiting for cluster [cpinjani-akswest (id: c-4zdpb)] to update node pool [pool11]"
time="2024-12-12T08:17:55Z" level=info msg="Checking configuration for cluster [cpinjani-akswest (id: c-4zdpb)]"
time="2024-12-12T08:17:56Z" level=info msg="Cluster [cpinjani-akswest (id: c-4zdpb)] finished updating"
time="2024-12-12T08:17:57Z" level=info msg="Checking configuration for cluster [cpinjani-akswest (id: c-4zdpb)]"
time="2024-12-12T08:17:59Z" level=info msg="Configuration for cluster [cpinjani-akswest (id: c-4zdpb)] was verified"
```

**PR's:**
- https://github.com/rancher/rancher/pull/48672
- https://github.com/rancher/rancher/pull/48716
- https://github.com/rancher/aks-operator/pull/836
- https://github.com/rancher/aks-operator/pull/830

cpinjani:
**Test Result**
Rancher version: `v2.11-1d345932e5f78b3da07ebe13691c8224c62b6240-head`, AKS Operator version: `v1.11.0-rc.3`

Issue is reproducible on above build. Also the full validation is blocked by follow-up issue https://github.com/rancher/aks-operator/issues/850

cc: @mjura @mitulshah-suse

[...]

PR #830, opened 2025-01-17
Check support for avaialibilty zones in region

mjura:
Issue: https://github.com/rancher/aks-operator/issues/751

Issue #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests
- [ ] backport needed


pkg/aks/check.go
```diff
@@ -175,3 +177,62 @@ func generateUniqueLogWorkspace(workspaceName string) string {
...
+	"westeurope":         true,
+	"westus2":            true,
+	"westus3":            true,
+}
+
+func CheckAvailabilityZonesSupport(location string) bool {
+	return azRegionSupport[location]
+}
+
+func CheckAvailabilityZonesSupportAPI(ctx context.Context, client services.SubscriptionsClientInterface, subscription string, location string) bool {
```
anmazzotti, line 220: Is this used? Because I see in the code the static map is tested instead, but I think the API way is better and the static list of locations can be removed.

mjura: I am not sure if API way is better in this case, because it will slow down aks-operator, because it will have iterate though all pages. From other side we have static map which is extremely easy and fast to check. The reason why I added static function is that new DC in new Regions are not changing at all. My favorite solution is to and use static map for aks-operator and eventually keep API func for future usage.

anmazzotti: I see, but then consider the list is going to eventually mutate and it seems to be longer in [this](https://gist.github.com/ausfestivus/04e55c7d80229069bf3bc75870630ec8?permalink_comment_id=5146240#gistcomment-5146240) discussion.

anmazzotti: And by the way I think it's also acceptable to read the list from the API at boot and cache it, it's not going to change that often anyway.

salasberryfin (approved):

mjura:
It was solved in https://github.com/rancher/aks-operator/pull/836

PR #837, opened 2025-01-31
[main] handle conflict errors while updating aksClusterConfig

krunalhinguu:

conflict errors occur since rancher also modifies aksconfig object. As conflict error returns empty object, errors like "Error recording akscc [] failure message: resource name may not be empty" occurs when we try to add a failure message to aksconfig status and update it.
The operation which results in conflict gets applied automatically in the next reconciliation call of the controller hence it is safe to just log it.

Issue  https://github.com/rancher/aks-operator/issues/665

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests
- [x] backport needed


vardhaman22:
superseded by https://github.com/rancher/aks-operator/pull/849

PR #863, opened 2025-02-19
[main] Prevent the cluster status from becoming active prematurely

krunalhinguu:

While updating multiple fields, changes were being applied one by one. During this process, the system periodically polls the cluster for status updates. However, the cluster status was prematurely transitioning to active instead of updating. This led to:
- Some updates being reverted.
- Certain changes getting skipped intermittently.

To address this, we now ensure that while making updates, the cluster status is enqueued as updating. This keeps the cluster continuously checking for upstream changes and prevents unintended timeouts.

[Issue #667](https://github.com/rancher/aks-operator/issues/667)

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests
- [x] backport needed

### DIFF
--- a/controller/aks-cluster-config-handler.go
+++ b/controller/aks-cluster-config-handler.go
@@ -853,10 +853,6 @@ func (h *Handler) updateUpstreamClusterState(ctx context.Context, config *aksv1.
 	}
 
 	if updateAksCluster {
-		// If status is not updating, then enqueue the update ( to re-enter the onChange handler )
-		if config.Status.Phase != aksConfigUpdatingPhase {
-			return h.enqueueUpdate(config)
-		}
 		resourceGroupExists, err := aks.ExistsResourceGroup(ctx, h.azureClients.resourceGroupsClient, config.Spec.ResourceGroup)
 		if err != nil && strings.Contains(err.Error(), "unauthorized") {
 			logrus.Infof("User does not have permissions to access resource group [%s]: %s", config.Spec.ResourceGroup, err)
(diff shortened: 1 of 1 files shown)


vardhaman22 (approved):
LGTM

controller/aks-cluster-config-handler.go
```diff
@@ -849,6 +853,10 @@ func (h *Handler) updateUpstreamClusterState(ctx context.Context, config *aksv1.
 	}
 
 	if updateAksCluster {
+		// If status is not updating, then enqueue the update ( to re-enter the onChange handler )
+		if config.Status.Phase != aksConfigUpdatingPhase {
+			return h.enqueueUpdate(config)
+		}
```
yiannistri, line 859: Technically, this check can be moved down to line 866 where the resource group is created if it doesn't exist, no? Since until that point, no other change is being made that would require the status to be set to `Updating`.

krunalhinguu: sure. I have moved the code. Thanks

controller/aks-cluster-config-handler.go
```diff
@@ -862,6 +866,11 @@ func (h *Handler) updateUpstreamClusterState(ctx context.Context, config *aksv1.
 			logrus.Infof("Resource group [%s] updated successfully", config.Spec.ResourceGroup)
 		}
 
+		// If status is not updating, then enqueue the update (to re-enter the onChange handler)
+		if config.Status.Phase != aksConfigUpdatingPhase {
+			return h.enqueueUpdate(config)
+		}
+
```
yiannistri, line 873: Should this block be inside the `if !resourceGroupExists` block a few lines up? Apologies if my previous review comment was confusing. In my opinion, we should enqueue an update right before we begin making any changes, does that make sense?

yiannistri (commented):
The suggested change looks ok to me. I still need to test this locally with multiple changes to see how it behaves.

yiannistri (approved):
Tested scenario described in bug and it worked as expected.

Issue #850, opened 2025-02-06
Unable to add Nodepool with no availability zone


cpinjani:
**Setup**
Rancher version: `v2.11-1d345932e5f78b3da07ebe13691c8224c62b6240-head`
AKS Operator version: `v1.11.0-rc.3`

**Description**
Adding Nodepool to AKS cluster without selecting any availability zone resulting into error: 
`Changing availability zones for node pool [pool1] in cluster [cpinjani-west (id: c-qlkpq)] is not permitted` 

**Steps to reproduce:**
- Create a cluster in region with AZ support Eg. EastUS
- Wait for cluster to be active and Edit cluster
- Add nodepool and do not select any availability zone

![Image](https://github.com/user-attachments/assets/36fd7fab-58f5-42d1-821a-a0286e9ba734)

vivek-shilimkar:
Validated on rancher v2.12-3a11b40d6d60d17279b22eaf9aedcfcebe80c6d4-head
 
- Created AKS v1.32.0 cluster
- Once cluster came into Active state added another node pool without any availability zone.
- Cluster came into Active state with a new node pool.

PR #913, opened 2025-03-27
[main] improve availability zone comparison logic

krunalhinguu:
### Issue:
Previously, adding a NodePool with no availability zone` (AvailabilityZones == nil)` was failing due to an issue with `reflect.DeepEqual()`. The method was not handling nil and empty slices correctly, causing unexpected mismatches between `np.AvailabilityZones` and `upstreamNodePool.AvailabilityZones`.

### Fix:
We replaced `reflect.DeepEqual()` with a custom function, `availabilityZonesEqual()`, which properly handles:

- Nil vs. empty slice comparisons
- Cases where both values are nil
- Ensuring element-wise comparison for non-nil slices

### Related

- https://github.com/rancher/aks-operator/issues/850

### DIFF
--- a/controller/aks-cluster-config-handler.go
+++ b/controller/aks-cluster-config-handler.go
@@ -969,8 +969,8 @@ func (h *Handler) updateUpstreamClusterState(ctx context.Context, config *aksv1.
 					logrus.Debugf("config: %s; upstream: %s", np.Mode, upstreamNodePool.Mode)
 					updateNodePool = true
 				}
-				if np.AvailabilityZones != nil && !reflect.DeepEqual(np.AvailabilityZones, upstreamNodePool.AvailabilityZones) {
-					return config, fmt.Errorf("Changing availability zones for node pool [%s] in cluster [%s (id: %s)] is not permitted", aks.String(np.Name), config.Spec.ClusterName, config.Name)
+				if np.AvailabilityZones != nil && upstreamNodePool.AvailabilityZones != nil && !availabilityZonesEqual(*np.AvailabilityZones, *upstreamNodePool.AvailabilityZones) {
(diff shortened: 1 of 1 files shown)


controller/aks-cluster-config-handler.go
```diff
@@ -1069,3 +1069,39 @@ func (h *Handler) getAzureClients(config *aksv1.AKSClusterConfig) error {
 
 	return nil
 }
+
+func availabilityZonesEqual(a, b *[]string) bool {
```
mjura, line 1073:
```suggestion
func availabilityZonesEqual(a, b []string) bool {
	// Consider nil slices and empty slices as equivalent
	if (a == nil || len(a) == 0) && (b == nil || len(b) == 0) {
		return true
	}
	
	// If lengths don't match, they're not equal
	if len(a) != len(b) {
		return false
	}
	
	// Compare each element
	for i := range a {
		if a[i] != b[i] {
			return false
		}
	}
	return true
}
```

mjura: this is much better

mjura (commented):
Takes slices directly instead of pointers to slices

controller/aks-cluster-config-handler.go
krunalhinguu: make sense

mjura: You are doing nil checks before calling this func

```
if np.AvailabilityZones != nil && upstreamNodePool.AvailabilityZones != nil && !availabilityZonesEqual(*np.AvailabilityZones, *upstreamNodePool.AvailabilityZones) {

```

so it doesn't make sense to me

krunalhinguu: make sense. I have updated the function @mjura

mjura (approved):

vardhaman22 (approved):

Issue #1516, opened 2026-07-08
Deprecate github.com/Azure/go-autorest


aruiz14:
This dependency was [deprecated in March 2023](https://github.com/Azure/go-autorest#note-support-for-modules-in-this-repo-ended-on-march-31-2023-additional-information-can-be-found-here) (archived on Feb 2025), but we are still importing it for some residual usage.
Given that `rancher/rancher` also imports `aks-operator`, it also gets added to rancher as a transitive dependency.

PR #1517, opened 2026-07-08
Remove github.com/Azure/go-autorest dependency

aruiz14:

This dependency was [deprecated in March 2023](https://github.com/Azure/go-autorest#note-support-for-modules-in-this-repo-ended-on-march-31-2023-additional-information-can-be-found-here) (archived on Feb 2025), but we are still importing it for some residual usage.
Given that rancher/rancher also imports aks-operator, it also gets added to rancher as a transitive dependency.

Issue #1516

The API diff check fails for obvious reasons:
```
github.com/rancher/aks-operator/pkg/utils
  Incompatible changes:
  - RequestWithInspection: removed
  - ResponseWithInspection: removed

github.com/rancher/aks-operator/pkg/aks
  Incompatible changes:
  - FindTenantID: changed from func(context.Context, github.com/Azure/go-autorest/autorest/azure.Environment, string) (string, error) to func(context.Context, string, string) (string, error)
  - GetEnvironment: changed from func(string) (github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud.Configuration, github.com/Azure/go-autorest/autorest/azure.Environment) to func(string) github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud.Configuration
  - NewSubscriptionsClient: removed
```
But I didn't find any external usage for those functions.

**Checklist**:

- [X] squashed commits into logical changes
- [ ] includes documentation
- [ ] adds unit tests
- [X] backport needed: probably, at least to `release-v2.15`


pkg/aks/client.go
```diff
@@ -88,9 +86,9 @@ func GetSecrets(_ wranglerv1.SecretCache, secretClient wranglerv1.SecretClient,
 	cred.SubscriptionID = string(subscriptionIDBytes)
 	cred.ClientID = string(clientIDBytes)
 	cred.ClientSecret = string(clientSecretBytes)
-	cred.Cloud = cloud
-	cred.AuthBaseURL = &env.ActiveDirectoryEndpoint
-	cred.BaseURL = &env.ResourceManagerEndpoint
+	cred.Cloud = env
+	cred.AuthBaseURL = new(env.Services[cloud.ResourceManager].Endpoint)
```
Abhishek-Valaboju: AuthBaseURL and BaseURL look swapped compared to the previous implementation. Could you please confirm if this is intentional.

pkg/aks/client.go
```diff
@@ -148,61 +146,55 @@ func GetCachedTenantID(secretClient secretClient, subscriptionID string, secret
...
+		return "", err
+	}
 
 	// we expect this request to fail (err != nil), but we are only interested
 	// in headers, so surface the error if the Response is not present (i.e.
 	// network error etc)
-	subs, err := c.Get(goCtx, subscriptionID)
-	if subs.Response.Response == nil {
-		return "", fmt.Errorf("Request failed: %v", err)
+	res, err := http.DefaultClient.Do(req)
```
Abhishek-Valaboju, line 182: With the switch to http.DefaultClient, is there any expected behavior change compared to the previous implementation?

aruiz14: It should continue to work. You can even use `curl` locally:
```
❯ curl -SsLi 'https://management.azure.com/subscriptions/XXXXXXXXX?api-version=2022-12-01' | grep www-auth
www-authenticate: Bearer authorization_uri="https://login.windows.net/YYYYYYYYYYYYYYYY", error="invalid_token", error_description="The authentication failed because of missing 'Authorization' header."
```

The main reason for this change is that creating a [SubscriptionsClient](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armsubscriptions/v2#NewClient) now requires to pass credentials. There is an alternative to use a lower-level client base client and something called "pipelines", but it all felt too complex and equally "hacky".

pkg/aks/client.go
aruiz14: Oh, nice catch. I've been applying this change in other places where they were in a different order, my bad.
Fixed in 3a9cf44

Abhishek-Valaboju (approved):

